PR #53, opened 2016-01-17
Remove `Route.prototype.mount`.

rwjblue:
Per discussion at core team meeting, removing `this.mount` from within an `Ember.Route`. In general, manual usage of `this.mount` can be replaced by more declarative usage of `{{mount}}`.


bugant:
@rwjblue does it means that routable engines will be sunset?

dgeb:
@bugant no, both routable and route-less engines will continue to be supported.

This PR simply removed one method of mounting route-less engines from within a `Route`. Instead we'll just support mounting route-less engines in templates via `{{mount}}`.

PR #60, opened 2016-02-02
Add in-repo-engine blueprint

trentmwillis:
I wanted to pitch in by adding some blueprints to help people get started using this addon.

Currently, this just uses the `in-repo-addon` blueprint and adds a `addon/engine.js` file. Other features I think this should include:

- [x] option to generate a routable or route-less engine
  - [x] routable engines should include `addon/routes.js`
  - [x] routable engines should modify `router.js`
  - [x] routeless engines should include `templates/application.hbs`


blueprints/in-repo-engine/files/lib/__name__/index.js
```diff
@@ -0,0 +1,8 @@
+/*jshint node:true*/
+module.exports = {
+  name: '<%= dasherizedModuleName %>',
+
+  isDevelopingAddon: function() {
+    return true;
+  }
+};
```
dgeb: Please update to use the new `EngineAddon` extension:

```
/*jshint node:true*/
var EngineAddon = require('ember-engines/lib/engine-addon');

   name: '<%= dasherizedModuleName %>',

  isDevelopingAddon: function() {
    return true;
  }
});
```

dgeb:
@trentmwillis seems like a good start - thanks!

I agree that routable vs route-less would be good options. The former could create `routes.js` and the latter could create `templates/application.hbs`.

blueprints/in-repo-engine/files/lib/__name__/index.js
rwjblue: We also had to disable hinting for the in-repo engines in this repo (see https://github.com/dgeb/ember-engines/blob/master/tests/dummy/lib/ember-chat/index.js#L11) to prevent errors from being thrown. Can you test if that is needed here also (I'm hoping it is not needed, but there may be bugs to fix upstream in ember-cli)?

trentmwillis: Will update. When I was working locally I didn't notice any hinting errors being thrown, so it might have been fixed with more recent dependencies. Will double check that though.

trentmwillis:
@dgeb do you know of the best way to accomplish that? As far as my understanding of blueprints goes, there isn't really a way to specify optional files. My thought is to just generate both and then remove the unused one in the `afterInstall` hook.

rwjblue:
You can implement `filePaths` and have separate sets of files for each, based on an option.

See https://github.com/ember-cli/ember-cli/pull/4837 and https://github.com/offirgolan/ember-cp-validations/pull/8 for some background.

trentmwillis:
@rwjblue awesome! Exactly what is needed. Will update this later today.

blueprints/in-repo-engine/files/lib/__name__/index.js
trentmwillis: @rwjblue from my experiments, it seems like the hinting errors only get thrown if you are using an actual addon (or symlinking an in-repo-addon as is done in the ember-engines tests). When using a straight up in-repo-addon as generated by default, it works without any issues.

rwjblue: @trentmwillis - Awesome, thanks for confirming this.  I think that in-repo-addons inside an addon should generally work without extra fuss (likely a bug in Ember CLI itself).

trentmwillis:
@rwjblue @dgeb I think this is ready for actual review now. There's a good bit of relying on behavior from other blueprints which I am unsure if this is the best approach or not. There's also a `TODO` that I'll probably open an issue for on Ember-CLI, but otherwise it works as expected.

rwjblue:
That TODO is exactly what https://github.com/ember-cli/ember-cli/issues/5481 is about.  I'd love help fixing that upstream (I believe I left a comment on potential direction in that issue).

blueprints/in-repo-engine/routable-files/lib/__name__/addon/routes.js
```diff
@@ -0,0 +1,3 @@
+export default function() {
```
rwjblue: I don't want to block this from being merged as I believe it will help many people add engines to their apps, but I'd like it if we could fix https://github.com/dgeb/ember-engines/issues/64 before releasing a version with this blueprint.

rwjblue:
This LGTM, I'd like to address https://github.com/dgeb/ember-engines/issues/64 in general but that can happen after we land this if you'd prefer.

@dgeb - Any objections?

dgeb:
Thanks @trentmwillis!

PR #61, opened 2016-02-02
Introduce `EngineAddon` extension.

dgeb:
The EngineAddon extension ensures that an engine is able to access
addon dependencies as if it is an application. This is accomplished
by merging each addon’s `app` tree into the engine’s module
namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,18 @@ described above.
 
 ### Configuring your Engine
 
+Within your engine's root directory, modify `index.js` so that your addon
+is configured as an engine using the `EngineAddon` extension:
+
+```js
+/*jshint node:true*/
+var EngineAddon = require('ember-engines/lib/engine-addon');
+
+module.exports = EngineAddon.extend({
+  name: 'ember-blog'
+});
+```
+
 Within your engine's `addon` directory, add a new `engine.js` file:
 
 ```js
--- a/bin/install-test-addons.sh
+++ b/bin/install-test-addons.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+mkdir -p node_modules
+
+rm -rf node_modules/common-components
+ln -s ../tests/dummy/lib/common-components node_modules/common-components
+
 rm -rf node_modules/ember-blog
 ln -s ../tests/dummy/lib/ember-blog node_modules/ember-blog
 
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1,0 +1,48 @@
+var Funnel = require('broccoli-funnel');
+var mergeTrees = require('broccoli-merge-trees');
+
+module.exports = {
+  extend: function(options) {
+    if (options.treeFor) {
+      throw new Error('Do not provide a custom `options.treeFor` with `EngineAddon.extend(options)`.');
+    }
+
+    options.treeFor = function treeFor(name) {
+      var tree, trees;
+
+      this._requireBuildPackages();
+
+      if (name === 'app') {
+        trees = [];
+      } else {
+        trees = this.eachAddonInvoke('treeFor', [name]);
+
+        if (name === 'addon') {
+          var appTree = mergeTrees(this.eachAddonInvoke('treeFor', ['app']), {
+            overwrite: true,
+            annotation: 'Engine#treeFor (' + options.name + ' - ' + name + ')'
+          });
+          var funneledAppTree = new Funnel(appTree, {
+            destDir: 'modules/' + options.name
+          });
+          trees.push(funneledAppTree);
+        }
+      }
+
+      if (tree = this._treeFor(name)) {
+        trees.push(tree);
+      }
+
+      if (this.isDevelopingAddon() && this.hintingEnabled() && name === 'app') {
+        trees.push(this.jshintAddonTree());
+      }
+
+      return mergeTrees(trees.filter(Boolean), {
+        overwrite: true,
+        annotation: 'Engine#treeFor (' + options.name + ' - ' + name + ')'
+      });
+    };
+
+    return options;
+  }
+}
--- a/package.json
+++ b/package.json
@@ -26,6 +26,8 @@
   "license": "MIT",
   "devDependencies": {
     "broccoli-asset-rev": "^2.2.0",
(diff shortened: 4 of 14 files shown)


trentmwillis:
I was looking into solving this same problem today and considered this approach. However, I am concerned that this will result in an explosion of modules and file size as it will be creating a new module for each shared module in each engine it is imported in to.

Other than it being a bit cumbersome, why not take a similar declarative approach as is currently working for services?

rwjblue:
> I am concerned that this will result in an explosion of modules and file size as it will be creating a new module for each shared module in each engine it is imported in to.

I'm not sure how this introduces very much duplication. If two things (EngineA and AppB) include the same addon (AddonC), the `addon` tree of AddonC is included into the resulting build once and the `app` tree is merged with AppB and EngineA.  Since the `app` trees of addons are simply modules with `export { default } from 'addon-c/whatever';` the actual implementation (`addon` tree) is shared and included once there are very few duplicated bytes even in the case of a shared addon.

Also, this only affects the direct addons listed in an engine's `dependencies`, not all addons in the entire consuming project/engine.

> Other than it being a bit cumbersome, why not take a similar declarative approach as is currently working for services?

That channel is for allowing things from the consuming application to be accessed by an engine, but has nothing to do with allowing an engine itself to use/consume addons of its own.

rwjblue:
@trentmwillis - I'm going to land this in the interest of having a workable solution here, I would absolutely like to continue discussing pros/cons and alternative implementations.

trentmwillis:
> I'm not sure how this introduces very much duplication.

@rwjblue after thinking about it more, it might be that this won't introduce much duplication, but I think that largely depends on developers doing the right thing.

If a developer hasn't done the right thing and has instead defined the entirety of the module in their addon's `app` directory, the entire module will be cloned. Perhaps that can be solved by simply auto-generating the import modules instead of cloning from the addon's app tree? Basically ensuring the module does the right thing.

Additionally, if addons are not granular enough, you wind up creating modules that aren't even used by the consuming Engine. On the developer end, maybe they need to breakup the addon, but that can become cumbersome if you have many shared modules (e.g., in the case of a UI Component addon).

> but has nothing to do with allowing an engine itself to use/consume addons of its own.

I understand this. I was just wondering if we should explore more declarative options to prevent issues such as creating import modules for unused modules (mentioned above) or potential collisions (ex: `addon-c/app/components/dropdown` and `engine-a/addon/components/dropdown`, both become `engine-a/components/dropdown`).

Instead of a declaration that takes effect at run-time (like we do for services) maybe it should be a build time declaration? We could default to including all modules with options to filter out modules, only include specific modules, and rename specific modules.

dgeb:
> If a developer hasn't done the right thing and has instead defined the entirety of the module in their addon's app directory, the entire module will be cloned.

I understand the concern, but I'd say the same is true for app's using addons today. We aren't dealing with a new problem here, just one that might be magnified by engines.

>  I was just wondering if we should explore more declarative options to prevent issues such as creating import modules for unused modules

Before this PR, this was the default behavior for engines. You could simply avoid using the `EngineAddon` extension and then directly import and re-export any modules from addons that you wish to use within an engine's `addon` dir. This approach gives you fine-grained declarative control and even allows for renaming, but is arguably pretty heavyweight and non-intuitive.

I think we should continue to explore options in other issues (since this is now merged) because I think we're all interested in striking the right balance.

rwjblue:
> If a developer hasn't done the right thing and has instead defined the entirety of the module in their addon's app directory, the entire module will be cloned.

Agreed, but this is no different than anything else.  We provide the patterns and easy path for folks to follow, but we cannot make them write good applications/addons/etc.  You what they say about leading a :horse: to :potable_water:...

> I was just wondering if we should explore more declarative options to prevent issues such as creating import modules for unused modules (mentioned above) or potential collisions (ex: addon-c/app/components/dropdown and engine-a/addon/components/dropdown, both become engine-a/components/dropdown).

This is exactly the same feature that exists for apps today. In this case `engine-a/addon/components/dropdown` should "win".  The end consumer (app or engine) should have the ability to override things provided in `app` trees from its addons.  This is an important extension point when consuming addons.

> Instead of a declaration that takes effect at run-time (like we do for services) maybe it should be a build time declaration? We could default to including all modules with options to filter out modules, only include specific modules, and rename specific modules.

Adding some API to allow an Engine to opt out or filter in some way the app tree of addons it is consuming seems totally fine (though I have no initial ideas on what that API should look like).

trentmwillis:
> but we cannot make them write good applications/addons/etc.

True, but could we introduce tooling to help ensure they're as close as possible? Going back to my above idea, would it be wrong to basically generate modules that do:

```
export { default } from 'app-that-tree-was-merged-into/components/foo-bar';
```

This way regardless of if the addon's module is simply an export or the entire module, we'll only have an export in the "duplicate" module.

> Adding some API to allow an Engine to opt out or filter in some way the app tree of addons it is consuming seems totally fine (though I have no initial ideas on what that API should look like).

Will chew on this a bit more and see if I can think of something reasonable.

rwjblue:
> `export { default } from 'app-that-tree-was-merged-into/components/foo-bar';`

The issue is that from the engine's perspective it has absolutely no knowledge if the addon that it is consuming is also consumed by something else above and/or parallel to it so there is no way to know at this level if it is a duplicate or not.

Also, what you are proposing is roughly the exact contents that are generated when you do `ember g component foo-bar` in an addon today.  It just reexports from the addon tree instead. So this would be roughly the same duplication that is already introduced (assuming that the addon developer followed the best practice of implementing in `addon/` and re-exporting in `app/`).

---

This may be much more possible when the linker/packager work is landed.  If we wanted to, it should be possible to detect duplicate module contents, and leverage the `define.alias` functionality in `loader.js`.

rwjblue:
> > Adding some API to allow an Engine to opt out or filter in some way the app tree of addons it is consuming seems totally fine (though I have no initial ideas on what that API should look like).

> Will chew on this a bit more and see if I can think of something reasonable.

RE: a filtering API; I think this belongs in Ember CLI itself.  The problems we are dealing with here are exactly the same as what is done in a normal consuming application's `app` tree, so the same general purpose API would apply.

PR #66, opened 2016-02-08
Make mount#rerender more robust.

dgeb:
Ensure that a controller is available before attempting to set its model. This fixes a scenario when a route-less engine is mounted alongside a routable engine, and then the routable engine is 
visited. We will need further testing to fully understand the problem that this fixes.


dgeb:
Merging to solve problems like #68. Still needs more investigation to understand the core issue.

Issue #69, opened 2016-02-11
In apps with legacy view support enabled: TypeError: Attempting to register an unknown factory: `view:select`


JackieChiles:
When `Ember.ENV._ENABLE_LEGACY_VIEW_SUPPORT` is on, the following error occurs, preventing the app from loading:

<pre>
TypeError: Attempting to register an unknown factory: `view:select`
</pre>

dgeb:
Fixed via #70 and published in v0.1.4

PR #70, opened 2016-02-11
Fix error when legacy view support is on.

JackieChiles:
When `Ember.ENV._ENABLE_LEGACY_VIEW_SUPPORT` is on, the following error occurs, preventing the app from loading:

<pre>
TypeError: Attempting to register an unknown factory: `view:select`
</pre>

This change to use Ember's dependency injection fixes the issue.

See issue #69.


dgeb:
Thanks!

PR #71, opened 2016-02-20
Quick hack to demonstrate linking between engines

diogomafra:
This is a quick hack to demonstrate linking between engines. It's motivated by this comment: https://github.com/emberjs/rfcs/pull/122#issuecomment-186496767

I think we can provide URLs for the engine in the same way we provide services. `url` is not the correct term, I'm using it here just because I could not think of a better name.

```js
// Engine declaration
export default Engine.extend({
  dependencies: {
    services: [
      'store',
    ],
    urls: [
      'new-user',
      'help'
    ]
  }
});
```

```js
// Application using that engine
App = Ember.Application.extend({
  engines: {
    emberBlog: {
      dependencies: {
        services: [
          'store'
        ],
        urls: [
          {'new-user': 'users.new'},
          {'help': 'help.blog'}
        ]
      }
    }
  }
});
```

```hbs
{{! links to external urls }}
{{#link-to get-url('new-user')}}Create user{{/link-to}}
{{#link-to get-url('help')}}Help{{/link-to}}
```

Note, this is just a hack to demonstrate my point in that RFC. Please don't merge. :smile:

### DIFF
--- a/addon/-private/engine-instance-ext.js
+++ b/addon/-private/engine-instance-ext.js
@@ -69,7 +69,7 @@ EngineInstance.reopen({
         let engineDependencies = engineConfiguration.dependencies;
 
         if (engineDependencies) {
-          ['services'].forEach((category) => {
+          ['services', 'urls'].forEach((category) => {
             if (engineDependencies[category]) {
               dependencies[category] = {};
               let dependencyType = this._dependencyTypeFromCategory(category);
@@ -215,7 +215,8 @@ EngineInstance.reopen({
       'route:basic',
       'event_dispatcher:main',
       '-bucket-cache:main',
-      'service:-routing'
+      'service:-routing',
+      'helper:get-url'
     ].forEach((key) => {
       this.register(key, parent.resolveRegistration(key));
     });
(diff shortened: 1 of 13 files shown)


krisselden:
Why is both get-url and link-to needed? Why not just get-url?

diogomafra:
> Why is both get-url and link-to needed? Why not just get-url?

Do you mean replacing the link-to component with get-url? (e.g. `{{#get-url 'new-user'}}Create{{/get-url}}`)

I don't think that's a good idea. We should be able to use the regular link-to, or [ember-href-to](https://github.com/intercom/ember-href-to), without creating a component just for linking outside the engine.

It might be possible to use regular links, without the get-url helper. In that case, the link-to component or the [Router Service](https://github.com/emberjs/rfcs/pull/95) would detect if this is an internal or external route. But this might make harder to understand where this route name is coming from. 

Note, this is just a hack to demonstrate my point in that RFC. Please don't merge. :smile:

krisselden:
No I meant '<a href={{get-url "new-user"}}>Create</a>'

diogomafra:
> If `get-url` returns a url, what is link-to for?

Sorry, my naming choice was really bad here. It doesn't return an url, it returns a route path.

knownasilya:
I like this approach, it's pretty obvious what urls are needed. This addresses the engine's need for specific urls, while the other approach (https://github.com/emberjs/rfcs/pull/122) is much looser, since there is no dependency, it's just arbitrary links in the engine that the author has to document, plus the consumer has to use the same routes in their app, which isn't always possible.

PR #72, opened 2016-02-23
Removes some whitespace to improve legibility

locks:


README.md
```diff
@@ -5,7 +5,6 @@ RFC](https://github.com/emberjs/rfcs/pull/10).
 
 Engines allow multiple logical applications to be composed together into a
 single application from the user's perspective.
-
```
dgeb: Maybe better to join the first two sentences together into a paragraph, as well as the following two, like so:

> This Ember addon implements the functionality described in the [Ember Engines
RFC](https://github.com/emberjs/rfcs/pull/10). Engines allow multiple logical applications to be composed together into a single application from the user's perspective.
>
> This addon must be installed in any ember-cli projects that function as either
consumers or providers of engines. The following functionality is supported:

dgeb: what do you think @locks ?

locks: sg, I can make the change if you want

dgeb: thanks @locks - I just merged and did a little housekeeping :)

Issue #57, opened 2016-01-26
Link out of an engine?


dfreeman:
Currently, all transition destinations in a routable engine are automatically relative to the engine's mount point, which makes perfect sense for the common case. As we're considering what our monolith might look like in an engines world, though, it's becoming clear that we have a set of core entities in our domain that most engines are going to want to be able to reference and potentially link out to.

I could imagine exposing the actual names of such engine-external routes via a service or some other dedicated configuration, and then just mocking out simple placeholder routes and pointing to those in the engine's dummy app during testing and isolated development.

To actually transition to these routes, though, we'd need a way to opt out of the automatic prefixing in  those specific `{{link-to}}` and `transitionTo()` invocations. Is this a use case that makes sense to support?

dgeb:
Thanks for raising this. There are definitely going to be situations in which engines will know exactly where they're being mounted, although this should not be the default case (as you point out).

@rwjblue and I have discussed the possibility of enabling links to parent routes through a custom `{{link-to-parent}}` helper (or similar). It's certainly technically possible, although we need to be careful to not encourage anti-patterns.

mike183:
This is something that I had been thinking about, but I think @dgeb's and @rwjblue's suggestion of something similar to a `{{link-to-parent}}` helper would be a better alternative as it would also make templates more explicit.

Something else that may also be worth considering, would the `{{link-to-parent}}` helper allow you to transition to routes from multiple levels up and not a direct parent? Or is that something that should be discouraged?

stefanpenner:
I would like to propose his be an independent RFC as many details need to be fleshed out

trentmwillis:
Would love to see this feature added as I think it fits a pattern we see frequently. Any application with a "tabbed" interface could potentially want each tab to be a separate Engine but would almost certainly have flows that would link between tabs/engines.

In that regard, I think `{{link-to-parent}}` is semantically confusing, does that mean you can only link to your parent application/engine? What about into a sibling engine? My initial thought is instead to name it `{{link-to-external}}` as in all cases you're going to something external to the Engine. But this is kind of a minor point.

Maybe I missed it, but is there any point in the RFC that outlines the expected behavior/scoping of links/routing within an Engine?

stefanpenner:
question that need answering in the RFC

how does one link to engines which have not yet been loaded ?

* how do QP support this case
* how do current/loading/active etc support this case.

nathanhammond:
[...]

PR #75, opened 2016-03-10
External links / linking between engines

trentmwillis:
This is a proof-of-concept for the recent updates to emberjs/rfcs#122 which attempts to address #57.

I would definitely like to see this feature move along, so I will update this branch with any future iterations on the RFC. This is last main feature I would need before using Engines in production projects.

### DIFF
--- a/addon/-private/engine-instance-ext.js
+++ b/addon/-private/engine-instance-ext.js
@@ -97,6 +97,10 @@ EngineInstance.reopen({
             }
           });
         }
+
+        if (engineDependencies.externalRoutes) {
+          dependencies.externalRoutes = engineDependencies.externalRoutes;
+        }
       }
 
       // Cache dependencies for child engines for faster instantiation in the future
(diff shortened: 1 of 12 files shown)


trentmwillis:
@rwjblue and @dgeb. Would appreciate it if y'all could take a look at this and the associated RFC (emberjs/rfcs#122).

As mentioned with @rwjblue at EmberConf, seems like the community is amenable to the proposed solution here and we could try it out in this addon as a proving grounds of sorts.

addon/-private/engine-instance-ext.js
```diff
@@ -230,21 +234,36 @@ EngineInstance.reopen({
     });
   },
 
+  getExternalRoute(routeName) {
```
rwjblue: Lets make this private for now...

rwjblue: Hmm, there are many other methods here without an `_` prefix.  I guess lets leave this for now, and make a new issue to underscore any new methods to these core classes...

rwjblue:
The changes here look good.  Can you also add details to the README for this?

addon/-private/engine-instance-ext.js
trentmwillis: I'll go ahead and prefix this, since the RFC for this is still in flight, we should probably err on the side of caution with the APIs we expose.

trentmwillis:
@rwjblue updated the README with some documentation

rwjblue:
Thanks @trentmwillis, LGTM!

dgeb:
Yes, thanks for all your work here and on the RFC, @trentmwillis! :shipit:

PR #81, opened 2016-04-13
Allow getting the modelFor application route

trentmwillis:
Currently, `modelFor` in an Engine does not allow you to get the model from the `application` route, this is a simple fix to allow that.

### DIFF
--- a/addon/-private/route-ext.js
+++ b/addon/-private/route-ext.js
@@ -62,7 +62,11 @@ Route.reopen({
 
     if (owner.routable) {
       let prefix = owner.mountPoint;
-      routeName = `${prefix}.${_routeName}`;
+      if (routeName === 'application') {
+        routeName = prefix;
+      } else {
+        routeName = `${prefix}.${_routeName}`;
+      }
     }
 
     return this._super(routeName, ...args);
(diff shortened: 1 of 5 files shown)


addon/-private/route-ext.js
```diff
@@ -62,7 +62,7 @@ Route.reopen({
 
     if (owner.routable) {
       let prefix = owner.mountPoint;
-      routeName = `${prefix}.${_routeName}`;
+      routeName =  _routeName !== 'application' ? `${prefix}.${_routeName}` : prefix;
```
rwjblue: Looks good, but I'd prefer to avoid the ternary if you don't mind.

```js
if (owner.routable) {
  let prefix = owner.mountPoint;
  if (routeName === 'application') {
    routeName = prefix;
  } else {
    routeName = `${prefix}.${_routeName}`;
  }
}
```

trentmwillis: Sure thing. Updated the PR.

rwjblue:
:+1: - LGTM

I think the tests are failing against current canary builds.  I know that @dgeb was looking into this a bit the other day, but I suspect more changes are needed.

dgeb:
> I think the tests are failing against current canary builds. I know that @dgeb was looking into this a bit the other day, but I suspect more changes are needed.

Yes, still a bit more to do. I'll rebase my changes over these.

PR #82, opened 2016-04-15
Remove reference to Ember.Select

andrewle:
`Ember.Select` was deprecated and then removed as of
emberjs/ember.js#13144. This removes ember-engine's reference to
`Ember.Select` and should also fix the issue reported here: https://github.com/dgeb/ember-engines/issues/76#issuecomment-208985230


andrewle:
@rwjblue :heart: :heart: :heart:

PR #88, opened 2016-04-20
Fix transitionTo in order to support the full API of the super class

prashn64:


rwjblue:
The change seems good, but can you give me an example of what this fixes (i.e. a code snippet that doesn't work before but does work after)?

trentmwillis:
This was breaking for QP-only transitions. So stuff like:

```js
this.transitionTo({ queryParams: { direction: 'asc' } });
```

Also, fixes `replaceWith` not being scoped it seems.

rwjblue:
Awesome, thanks for the explanation!

Issue #89, opened 2016-04-21
paramsFor does not work properly with queryParams on page load.


prashn64:
EDIT: Simplifying the use case

routes/search.js
  ```
    model(params) {
      return params; // params are empty, params should be { keywords: 'foo' }
    }
  ```

Go to /search?keywords=foo

I'm looking for the origin of the bug, but figured I'd post it if anyone has any ideas.

PR #90, opened 2016-04-21
Pass the correct params object passed to the invoked model hook

prashn64:
Issue #89 

Also fixed an issue where modelFor('...') was not returning the correct value when a transition was fully completed.

I don't think this is the best way to do this, would like some good feedback on this PR.  There are also a ton of places where a prefix based on the mount point is used, so may be a good idea to standardize at a higher level.


prashn64:
Any opinions on this?

Issue #85, opened 2016-04-18
Component fails to re-render when used with bound value in Engine


trentmwillis:
**Issue:** When using a component in an Engine (either routeless or routable as far as I can tell) with a bound value, if you try to re-render it by modifying the bound value, the re-render will fail with an error like:

```bash
Uncaught Error: Assertion Failed: A helper named '<component-name>' could not be found
```

**Potential Cause:** After a good amount of digging, it appears that when the re-render occurs the lookups are done in the context of the consuming application and not the engine. Thus, this error makes sense as the component won't be found and the application will fall back to trying to lookup a helper of the same name (which will also not be found).

Obviously, this is not the intended behavior and the re-render of a component should occur in the same context as the original render. However, my knowledge of the rendering internals in Ember is not very good, so I didn't make much progress on any actual solution.

**Repro:** I've put together a simple reproduction of the issue on the [`re-render` branch of my fork](https://github.com/trentmwillis/ember-engines/tree/re-render). If you start that up and visit the `routable-engine-demo/blog` route, the error will occur.

cc @nathanhammond @rileyhilliard

dgeb:
Thanks for reporting, @trentmwillis. I'll investigate.

PR #92, opened 2016-04-22
Fix environment owner during re-render

trentmwillis:
Addressing #85 to the best of my ability.

**Root Issue**: During re-render, the `env.owner` was not correct within an `outlet`. This was because `outlet.render` was not called, meaning that the check for ensuring the proper `owner` never occurred.

**Solution:** Move the check for `owner` into `outlet.childEnv`, which is called on initial and subsequent render.

I'm not 100% certain this is the best solution as I still don't fully understand the rendering flows and hooks, but it works. I've included a previously failing test as well.

cc @krisselden


krisselden:
Just to leave a comment on why, `childEnv` is always called, but the `render` hook is only called if the outlet is dirty.

PR #95, opened 2016-05-03
Fix test addons

### LINKED ISSUE #94: Depends on unpublished ember-chat package
Looks like a devDependency on ember-chat may have been added by mistake.

ef4:
This configures the test addons as proper in-repo addons so no symlink wrangling is necessary.

Closes #94.


rwjblue:
This is how we started as well, however in-repo-addons are also included by consumers of `ember-engines`. If we land this and publish a version all users of `ember-engines` would receive the error you described in #94. The current situation while not ideal, is still better than forcing errors on our consumers.

PR #96, opened 2016-05-03
Run initializers and instance-initializers inside engines

ef4:
This enables initializers and instance-initializers inside engines.

The only change necessary to get instance-initializers running is altering the canonical addon/engine.js to invoke ember-load-initializers (which is analogous to what happens in app/app.js).

Getting the initializers running required an extention to the Engine class. Applications run them when they're told to boot, but no equivalent step happens for child engines, so instead we run them the first time we build an engine instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,10 +100,8 @@ import Engine from 'ember-engines/engine';
 import Resolver from 'ember-engines/resolver';
 import loadInitializers from 'ember-load-initializers';
 
-let Eng;
 const modulePrefix = 'ember-blog';
-
-Eng = Engine.extend({
+const Eng = Engine.extend({
   modulePrefix,
   Resolver
 });
--- a/README.md
+++ b/README.md
@@ -98,12 +98,20 @@ Within your engine's `addon` directory, add a new `engine.js` file:
 ```js
 import Engine from 'ember-engines/engine';
 import Resolver from 'ember-engines/resolver';
+import loadInitializers from 'ember-load-initializers';
 
-export default Engine.extend({
-  modulePrefix: 'ember-blog',
+let Eng;
+const modulePrefix = 'ember-blog';
 
+Eng = Engine.extend({
+  modulePrefix,
   Resolver
 });
+
+loadInitializers(Eng, modulePrefix);
+
+export default Eng;
+
 ```
 
 It's important to define a `modulePrefix` that will be used to resolve your
--- a/addon/-private/engine-ext.js
+++ b/addon/-private/engine-ext.js
@@ -18,6 +18,20 @@ const {
 } = Ember;
 
 Engine.reopen({
+  _initializersRan: false,
+
+  ensureInitializers() {
+    if (!this._initializersRan) {
+      this.runInitializers();
+      this._initializersRan = true;
+    }
+  },
+
+  buildInstance(options) {
+    this.ensureInitializers();
+    return this._super(options);
+  },
+
   buildRegistry() {
     let namespace = this;
     let registry = this._super(...arguments);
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -98,12 +98,20 @@ Within your engine's `addon` directory, add a new `engine.js` file:
...
 import Engine from 'ember-engines/engine';
 import Resolver from 'ember-engines/resolver';
+import loadInitializers from 'ember-load-initializers';
 
-export default Engine.extend({
-  modulePrefix: 'ember-blog',
+let Eng;
+const modulePrefix = 'ember-blog';
 
+Eng = Engine.extend({
```
rwjblue: `export default const Eng = Engine.extend(`

README.md
```diff
@@ -98,12 +98,20 @@ Within your engine's `addon` directory, add a new `engine.js` file:
 ```js
 import Engine from 'ember-engines/engine';
 import Resolver from 'ember-engines/resolver';
+import loadInitializers from 'ember-load-initializers';
 
-export default Engine.extend({
-  modulePrefix: 'ember-blog',
+let Eng;
```
rwjblue: Can you tweak this to something like the following:

```js
const modulePrefix = 'ember-blog';
const Eng = Engine.extend({
  modulePrefix,
  Resolver
});

loadInitializers(Eng, modulePrefix);
```

Reading it top to bottom like this seems easier to my brain 😝

rwjblue:
>  Applications run them when they're told to boot, but no equivalent step happens for child engines

I do not think this is quite correct. We should be running `initializers` during `init` of the `Engine` instance (this is roughly what happens with `Ember.Application` also).

---

Can you another test that ensures all initializers are ran before instance initializers? I am also concerned about an initializer calling `deferReadiness` on the engine (which IMO we should not support but should provide a helpful error).

Can you also update the blueprints ([here](https://github.com/dgeb/ember-engines/blob/master/blueprints/in-repo-engine/routable-files/lib/__name__/addon/engine.js) and [here](https://github.com/dgeb/ember-engines/blob/master/blueprints/in-repo-engine/routeless-files/lib/__name__/addon/engine.js)) to follow the recommendations in the README?

README.md
ef4: Agreed, I was just blindly copying the format of the app/app.js blueprint. This is much better.

ef4:
> I do not think this is quite correct. We should be running initializers during init of the Engine instance (this is roughly what happens with Ember.Application also).

`Application` doesn't run them during init, it waits until `domReady()` or `boot()`.

rwjblue:
> Application doesn't run them during init, it waits until domReady() or boot().

You are correct, but they are called as a byproduct of calling `Ember.Application.create` (there is a bit async there due to the domReady checks).  Since we do not intend to support async like that, we should fire the initializers just after init is finished (perhaps `on('init', function() { this.runInitializers(); })`).

rwjblue:
Anyways, I don't think that solution really ends up being different than this.  And arguably this is clearer...

dgeb:
I am fine with this approach to Engine init for now, with the understanding that we will refine it when we introduce async loading of engines.

rwjblue:
Thanks @ef4!

ef4:
Oh, one small nitpick: this change technically makes every engine need to put ember-load-initializers into `dependencies`.  Which doesn't even work, due to the dreaded nested `included` issue.

But in practice every ember-cli app already has ember-load-initializers, so it seems OK.

rwjblue:
Yes, and they can easily not include it if they don't need/want initializers.

PR #97, opened 2016-05-03
Fix npm install for local development

trentmwillis:
Addressing #94.

Essentially, this tells NPM to install the addons from the local file system. We want to keep the `prepublish` script however, so that the addons still get symlinked to make changes appear more easily.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ before_install:
 
 install:
   - npm install -g bower
-  - npm run prepublish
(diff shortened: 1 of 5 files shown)


dgeb:
This is the combo I was trying - and you just beat me to it 👍 

> We want to keep the prepublish script however, so that the addons still get symlinked to make changes appear more easily.

Yep, this is the key to making development bearable.

PR #100, opened 2016-05-04
Fix paramsFor when being called on the application route

prashn64:


rwjblue:
This looks good to me, thank you!  

All of these nuances is making me want real tests.

PR #101, opened 2016-05-08
Routeable engine instances should be destroyed

ef4:
This change causes routeable engine instances and everything in their containers to be properly destroyed when an app is destroyed.

The routeless ones are also missing this feature, but I will file a separate issue for that, since I'm less clear on the right way to fix it.


rwjblue:
Thank you!

PR #109, opened 2016-05-12
Remove duplicate addRouteForEngine call

nlfurniss:
`addRouteForEngine` is called inside `EmberRouterDSL.mount`, before `EmberRouterDSL.push` is called. `push` also calls addRouteForEngine, so it seems that it doesn't need to be done twice.

### DIFF
--- a/addon/-private/router-dsl-ext.js
+++ b/addon/-private/router-dsl-ext.js
@@ -44,8 +44,6 @@ EmberRouterDSL.prototype.mount = function(_name, _options) {
   let localFullName = 'application';
   let routeInfo = assign({ localFullName }, engineInfo);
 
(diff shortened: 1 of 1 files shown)


addon/-private/router-dsl-ext.js
```diff
@@ -44,8 +44,6 @@ EmberRouterDSL.prototype.mount = function(_name, _options) {
   let localFullName = 'application';
   let routeInfo = assign({ localFullName }, engineInfo);
 
-  this.options.addRouteForEngine(fullName, routeInfo);
```
trentmwillis: On further inspection, this is needed. Notice that `routeInfo` contains `localFullName` which is `"application"`. Which is different from the `localFullName` used below.

nlfurniss:
Trent's right, this is fine

Issue #35, opened 2016-01-08
Add tests/confirm nested routeable engines function properly.


rwjblue:

trentmwillis:
@rwjblue if you could provide a few more details on what would need to be tested (e.g., define what "function properly" entails) here I'd gladly help out.

villander:
it's already fixed as well @dgeb

PR #110, opened 2016-05-13
fix for mounting engines within an engine

kevlu33:
With these changes we can do some enginception, or mounting an engine within an engine.

Not sure if this is the right place to put the fix, but when processing an engine's routes, calling mount again breaks because this.options.engineInfo points to the outer engine's engineInfo.

### DIFF
--- a/addon/-private/router-dsl-ext.js
+++ b/addon/-private/router-dsl-ext.js
@@ -33,12 +33,23 @@ EmberRouterDSL.prototype.mount = function(_name, _options) {
 
   let callback;
   if (engineRouteMap) {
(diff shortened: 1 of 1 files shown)


dgeb:
@kevlu33 Thanks for the PR. I suppose we'll need a new engine loaded into the dummy app in order to test this. Are you up for adding tests?

dgeb:
I'm going to merge this since we're already tracking the need to test mounting routable engines within each other in #35 

Thanks again @kevlu33

kevlu33:
Thanks for merging. I noticed this morning that I left out an un-added file, will post a new pull request.

PR #111, opened 2016-05-13
[Clean] Eliminate registrations moved upstream.

dgeb:
Remove registrations moved into ember core via https://github.com/emberjs/ember.js/pull/13459

Cleanup of patches moved upstream into ember core will be done on the `clean` branch so that `master` can continue to accept patches and work in a backward compatible way as much as possible. Once all the patches are moved upstream, we'll merge `clean` to `master` and bump the minor version.

### DIFF
--- a/addon/-private/engine-ext.js
+++ b/addon/-private/engine-ext.js
@@ -1,20 +1,11 @@
 import Ember from 'ember';
-import LinkComponent from './link-to-component';
+import EngineScopedLinkComponent from './link-to-component';
 import ExternalLinkComponent from './link-to-external-component';
 import emberRequire from './ext-require';
 
-const EmberView = emberRequire('ember-views/views/view');
-const RoutingService = emberRequire('ember-routing/services/routing');
 const Engine = emberRequire('ember-application/system/engine');
-const P = emberRequire('container/registry', 'privatize');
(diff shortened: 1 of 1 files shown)


addon/-private/engine-ext.js
```diff
@@ -34,73 +25,11 @@ Engine.reopen({
...
-      registry.injection('route', '_bucketCache', P`-bucket-cache:main`);
-      registry.injection('controller', '_bucketCache', P`-bucket-cache:main`);
-
-      registry.injection('route', 'router', 'router:main');
-
-      registry.register('component:-text-field', TextField);
-      registry.register('component:-text-area', TextArea);
-      registry.register('component:-checkbox', Checkbox);
-      // registry.register('view:-legacy-each', LegacyEachView);
       registry.register('component:link-to', LinkComponent);
```
rwjblue: Can you rename this one?  I keep getting confused if it is the normal `Ember.LinkComponent` or our custom scoped one.  Perhaps you can name the imported variable `EngineScopedLinkComponent`?

rwjblue:
Left one nit-pick/tweak (non-blocking), but looks good to me!

addon/-private/engine-ext.js
dgeb: Ah, good call. I get tripped up by that one too.

dgeb:
Thanks @rwjblue!

PR #112, opened 2016-05-13
part 2 of engine within engine fix

kevlu33:
Should have been part of https://github.com/dgeb/ember-engines/pull/110


dgeb:
Ok, thanks again @kevlu33

PR #114, opened 2016-05-15
[WIP] Add guides as more thorough documentation on Engines

trentmwillis:
Due to the number of questions/requests for documentation that have been received through various channels, I figured it would be good to have some more thorough documentation on Engines than what is available in the `README`.

I realize that we want to be careful about codifying the experimental parts of this addon, but I think a large portion of the features are ready to begin adoption and lots of developers would like to use them so it makes sense to have some more thorough reading available.

My goals for this guide are that it will:

1. Evolve with the addon,
2. Provide insight into not just "how to do something" but "why things are done the way they are",
3. Give a central reference point for information that is currently scattered throughout issues, rfcs, slack convos, etc., and
4. Allow easier adoption of Engines for users new to some advanced Ember concepts

In it's current structure, the guide reads like a tutorial and explains concepts and ideas along the way. I think this format makes the most sense to encourage developers to get a holistic view of what Engines involve rather than picking up bits and pieces as they need them.

_Note: this is still a work-in-progress, but I want to get some feedback about the format and content ahead of time._


dgeb:
@trentmwillis Thanks for getting this started! The content here is really solid 💯 

At this point, my only hesitation is _where_ to put all this good content. My inclination is to set up a jekyll site for this repo.

trentmwillis:
> My inclination is to set up a jekyll site for this repo.

We could do that. I had wondered if maybe we should do something like [FastBoot](http://www.ember-fastboot.com/), but since Engines are still very much beta (alpha?) software, I was thinking maybe just an in-repo guide similar to [Glimmer](https://github.com/tildeio/glimmer/tree/master/guides) would suffice.

I'll leave the decision up to you. I plan to finish fleshing out the content here this weekend.

dgeb:
Merged via ba5ea264c1ec4c69a714d3e038cb0cd6217efec9

@trentmwillis Sorry for letting this sit, and thanks again for the great work :shipit: 

I'll talk over goals for a docs site with the rest of the core team.

PR #127, opened 2016-06-07
streams dependency was moved into ember-htmlbars

nucleartide:
This fixes the issue mentioned by @thousand [here][1]. Maybe it'd be a good idea to refer to a specific Ember.js commit when installing Ember?

[1]: https://github.com/dgeb/ember-engines/issues/126#issuecomment-224102436

### DIFF
--- a/addon/-private/keywords/mount.js
+++ b/addon/-private/keywords/mount.js
@@ -2,7 +2,7 @@ import Ember from "ember";
 import emberRequire from '../ext-require';
 
 // const internal = emberRequire('htmlbars-runtime').internal;
-const read = emberRequire('ember-metal/streams/utils', 'read');
+const read = emberRequire('ember-htmlbars/streams/utils', 'read');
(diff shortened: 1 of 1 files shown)


nucleartide:
Nvm, I guess I'll just use Ember 2.6 beta as mentioned [here][1].

[1]: https://github.com/dgeb/ember-engines/issues/120#issuecomment-220756279

thousand:
I think your change is will be appropriate eventually, maybe when the current canary becomes beta. If this is going away from ember-metal, which is likely since it's in the canary, then this would be needed long term.

Cryrivers:
@nathanhammond mentioned an appropriate fix before, https://github.com/ember-a11y/ember-a11y/commit/8d35e3755f79d2d8c9249fe747f600f61a4fadb1

should do it in that way i bet.

rwjblue:
It is only a temp fix anyways. In the long run, ember-htmlbars package will be going away completely. The implementation of read is like 2 lines, if we need it we should just inline it.

PR #128, opened 2016-06-07
Support engine configuration settings.

dgeb:
Engines can now store their own configuration settings in 
`config/environment.js`. 

Adjust examples to import this module and use it to define 
`modulePrefix`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,14 +94,32 @@ module.exports = EngineAddon.extend({
 });
 ```
 
+Within your engine's `config` directory, create a new `environment.js` file:
+
+```js
+/*jshint node:true*/
+'use strict';
+
+module.exports = function(environment) {
+  var ENV = {
+    modulePrefix: 'ember-blog',
+    environment: environment
+  }
+
+  return ENV;
+};
+```
+
 Within your engine's `addon` directory, add a new `engine.js` file:
 
 ```js
 import Engine from 'ember-engines/engine';
-import Resolver from 'ember-engines/resolver';
+import Resolver from 'ember-engines/resolver'; // <=== IMPORTANT - custom resolver!!!
 import loadInitializers from 'ember-load-initializers';
+import config from './config/environment';
 
-const modulePrefix = 'ember-blog';
+const { modulePrefix } = config;
+
 const Eng = Engine.extend({
   modulePrefix,
   Resolver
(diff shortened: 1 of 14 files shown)


dgeb:
I'm PR'ing this first against the v0.2 branch, which is currently green, in preparation for a new v0.2.x release. I'll bring this back into `clean` and finally `master` once they stabilize.

README.md
```diff
@@ -94,27 +94,42 @@ module.exports = EngineAddon.extend({
...
+  return ENV;
+};
+```
+
 Within your engine's `addon` directory, add a new `engine.js` file:
 
 ```js
 import Engine from 'ember-engines/engine';
-import Resolver from 'ember-engines/resolver';
+import Resolver from 'ember-engines/resolver'; // <=== IMPORTANT - custom resolver!!!
```
rwjblue: LOL

README.md
```diff
@@ -94,27 +94,42 @@ module.exports = EngineAddon.extend({
...
 import loadInitializers from 'ember-load-initializers';
+import config from './config/environment';
 
-const modulePrefix = 'ember-blog';
 const Eng = Engine.extend({
-  modulePrefix,
+  modulePrefix: config.modulePrefix,
```
rwjblue: Maybe destructure like above, so we can reuse the `modulePrefix` just below?

rwjblue:
I think we need to update the blueprints with these changes too.

rwjblue:
A few other notes RE `engineConfig`:

* We should allow the user to specify an `engineConfig` function in their options hash (which would override the default).  This would be very similar to the `config` hook.
* We should probably invoke `config` on each of our dependencies (like [project.getAddonsConfig](https://github.com/ember-cli/ember-cli/blob/master/lib/models/project.js#L219-L231)).

dgeb:
@rwjblue Good call on the blueprints and destructuring. I've made those changes.

> We should allow the user to specify an engineConfig function in their options hash (which would override the default). This would be very similar to the config hook.

I think this should work now - `contentFor` [gets the engine's config](https://github.com/dgeb/ember-engines/blob/f6d67600484f905f39782bee84dada2ef8073053/lib/engine-addon.js#L65) from `engineConfig`

dgeb:
> We should probably invoke config on each of our direct dependencies (like project.getAddonsConfig).

Another good call. I just implemented this.

dgeb:
@rwjblue Thanks for the review. I think I addressed all of your initial concerns and suggestions.

PR #131, opened 2016-06-08
Update README.

dgeb:
* Include note about compatibility and stability.
* Include new section about accessing configuration settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,20 @@ Support for the following concepts is under consideration:
 * Sharing of dependencies other than services and route paths.
 * Passing configuration attributes from an engine's parent.
 
+## Important Note about Compatibility and Stability
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -253,6 +254,37 @@ this.transitionToExternal('settings');
 
 For further documentation on this subject, view the [Engine Linking RFC](https://github.com/emberjs/rfcs/pull/122).
 
+### Accessing Engine Configuration Settings
+
+As in an application, you can provide configuration settings for your
+engine in `addon/config/environment.js`. You can access these settings in a
```
rwjblue: I believe this path is off, it should be `<engine-root>/config/environment.js` (which you then import as if it were located in `addon/config/environment.js`).

PR #137, opened 2016-06-15
Adding support for custom configPath

### LINKED ISSUE #133: Allow custom config path in package.json using configPath
Instead of a "config" folder, our engine has "ember-config" as the config directory.

In package.json, we have this specified:
```
"ember-addon": {
  "configPath": "ember-config"
},
```

However, this value is ignored. My `config` object is returned as empty. When I rename "ember-config" to "config", my `config` object is resolved correctly.

cklimkowsky:
Fixes #133, allowing the config directory to be named something other than "config"


dgeb:
Thanks @cklimkowsky!

rwjblue:
Nice, thank you!

Issue #126, opened 2016-06-02
route-less engine "crashes" on component property mutation


thousand:
I'm developing a route-less engine. It has a main parent component that handles opening and closing some child components. That main parent component keeps the data for its children in an array property and shows one child for each element in the array. The biggest issue i discovered around this is that if I mutate that array, adding or removing an element causing a child component to be added or removed, the engine gets removed from the DOM entirely, with no warnings or errors in the console. Additionally the component lifecycle event handlers like `willUpdate` or `willDestroyElement` are called. Example code:

```js
// parent-component.js
import Ember from 'ember';
export default Ember.Component.extend({
  init() {
    this.openChildren = [{name: 'John'}, {name: 'Paul'}, {name: 'George'}]
  },

  willUpdate() {
    // i'm never called
    console.log('updating...');
  },

  willDestroyElement() {
    // i'm never called
    console.log('destroying...');
  },

  actions: {
    openChild(child = {name: "Ringo"}) {
      let openChildren = this.get('openChildren');
      openChildren.addObject(child);
    },

    closeChild(child) {
      let openChildren = this.get('openChildren');
      openChildren.removeObject(conversation);
    }
  }

});
```

```hbs
{! parent-component.hbs }
<h1>Parent</h1>

{{#each openChildren as |child|}}
  {{child
    child=child
    closeSelectedChild="closeChild"
  }}
{{/each}}
```

```js
// child-component.js
import Ember from 'ember';
export default Ember.Component.extend({
  actions: {
    close() {
      this.sendAction('closeSelectedChild', this.get('child'));
    }
  }
});
```

```hbs
{! child-component.hbs }
<header>
  <h2>{{name}}</h2>
  <button {{action="close"}}>X</button>
</header>
```

Simpler component property mutations are also not working, though without the catastrophic results. Just a simple iterator like `<button {{action "increment"}}>{{value}}</button>` also does not update, and throws this error in the console on the first mutation: `Uncaught Error: Assertion Failed: A helper named 'parent-component' could not be found`

trentmwillis:
I feel like the root issue here is probably similar to the one in #85.

@thousand could you put together a simple app to clone down so we can debug further? Or, even better, would be if you could set up a failing test in the dummy app inside this addon.

thousand:
[...]

PR #138, opened 2016-06-15
Refactoring mount keyword to properly build child environment in render flows.

thousand:
This is at least a proof of concept to address the problem identified in #126 to the best of my ability. Adding a prop onto `state` feels ugly to me, so please advise if there's a better place to put this functionality.

Steps taken: 
* Break engine instance setup into separate function
* Moved assertions to match where engine will be first instantiated
* Attach `engineInstance` to `state` where it can be re-referenced in subsequent renders of child nodes
* generate render environment in `childEnv` based on `engineInstance`'s `application` view and template.

### DIFF
--- a/addon/-private/keywords/mount.js
+++ b/addon/-private/keywords/mount.js
@@ -64,7 +64,7 @@ registerKeyword('mount', {
     }
   },
 
-  setupState(prevState, env, scope, params, hash) {
+  setupState(prevState, env, scope, params /*, hash */) {
     let name = params[0];
 
     assert(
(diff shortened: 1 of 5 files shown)


thousand:
@dgeb @rwjblue any thoughts here? If there's a better approach please let me know and i'll be happy to implement it.

dgeb:
@thousand sorry for the delay - I will be reviewing this very soon.

dgeb:
Included in #142. Thanks @thousand!

PR #145, opened 2016-06-25
Fix a typo

steveszc:
Structure was spelled wrong (sturcture) on line 13.


rwjblue:
Good catch, thank you!

PR #146, opened 2016-07-06
Ignore node_modules from test engines/addons

mike-north:
running 
```sh
./bin/install-test-addons.sh
```
leaves the user in a dirty git state. Because this is in the prepublish hook, it can happen when doing things like `npm link`.


rwjblue:
Thanks @mike-north!

PR #147, opened 2016-07-06
[v0.2] Get route blueprint from ember-legacy-blueprints for ember-cli 2.6+

mike-north:
**Note: this branch is based on the v0.2 branch**


.gitignore
```diff
@@ -15,3 +15,4 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
+tests/dummy/lib/**/node_modules
```
mike-north: Same as merged PR: #146

blueprints/in-repo-engine/index.js
```diff
@@ -32,14 +34,31 @@ module.exports = {
...
+    this.shouldTouchRouter = this.routeBlueprint.shouldTouchRouter;
+    this.shouldEntityTouchRouter = this.routeBlueprint.shouldEntityTouchRouter;
+  },
+
   install: function(options) {
     this.options = options;
-    return this._super.install.apply(this, arguments);
+    var superResult = this._super.install.apply(this, arguments);
+    this._setupRouteBlueprint();
+    return superResult;
```
mike-north: This approach is a little awkward, but I see no other way given that:
* We have to consider that super's install hook may return a promise, and so we must pass that through
* `this._super.install` (and uninstall) have side effects that `ember-cli-version-checker` depend on (specifically, the existence of the `project` attribute, from which the npm dependency data is fetched)

package.json
```diff
@@ -35,10 +35,12 @@
     "ember-cli-htmlbars-inline-precompile": "^0.3.1",
     "ember-cli-inject-live-reload": "^1.4.0",
     "ember-cli-jshint": "^1.0.0",
+    "ember-cli-legacy-blueprints": "0.1.1",
```
mike-north: If necessary, we could add conditional installation of this dependency to the `ember-engines` installation blueprint (i.e., only for ember-cli >= 2.6), but this could cause some tricky unexpected failures as developers upgrade from 2.5 --> 2.6 and don't understand why their in-repo-engine generator started to fail

dgeb:
Thanks for your work here @mike-north!

Given that the `v0.2` branch is intended to only support ember v2.6.x, I have no problem requiring ember-cli v2.6.x as well for the next v0.2.x release.

> If necessary, we could add conditional installation of this dependency to the ember-engines installation blueprint (i.e., only for ember-cli >= 2.6), but this could cause some tricky unexpected failures as developers upgrade from 2.5 --> 2.6 and don't understand why their in-repo-engine generator started to fail
> Of course this could also be addressed by throwing appropriate errors upon execution of the blueprint.

Yes, I think that throwing errors when the blueprint is installed would be appropriate here.

Please let me know if you need additional feedback to wrap this up. Thanks again!

mike-north:
> Given that the v0.2 branch is intended to only support ember v2.6.x, I have no problem requiring ember-cli v2.6.x as well for the next v0.2.x release.

There's nothing preventing this fix from being applied to master. This just happens to be the branch I'm using with ember-cli 2.6 for our apps that make use of engines. Totally up to you as to what's best for the addon.

> Yes, I think that throwing errors when the blueprint is installed would be appropriate here.

understood. will add this

blueprints/in-repo-engine/index.js
```diff
@@ -32,14 +34,39 @@ module.exports = {
...
+    var checker = new VersionChecker(this);
+    var emberCliVersion = checker.for('ember-cli', 'npm');
+
+    var routeBlueprintPkg = emberCliVersion.isAbove('2.6.0')
+      ? OLD_ROUTE_BLUEPRINT_PKG
+      : NEW_ROUTE_BLUEPRINT_PKG;
+
+    if (emberCliVersion.isAbove('2.6.0')) {
+      var legacyBlueprintsVersion = checker.for('ember-cli-legacy-blueprints', 'npm');
+      if (!legacyBlueprintsVersion.version) {
```
mike-north: per https://github.com/rwjblue/ember-cli-version-checker/issues/8#issuecomment-115843950

mike-north:
Please reach out if there's anything else this PR needs. AFAIK, it's the missing piece to allow a "cutting edge" ember setup (fastboot, engines, ember-cli#beta, ember#beta, etc...) to work seamlessly with published releases.

dgeb:
@mike-north just giving it a run through now ... hope to merge soon.

dgeb:
@mike-north looking pretty good. I think we actually need to move `ember-cli-legacy-blueprints` and `ember-cli-version-checker` from devDependencies to dependencies, since they're required by projects that use `ember-engines`. Would you be up for PR'ing this change to `v0.2` and bringing all these changes back into `master` as well?

dgeb:
See https://github.com/dgeb/ember-engines/pull/151 for a simplified solution.

PR #152, opened 2016-07-12
Allow custom engine config locations via `engineConfigPath`.

### LINKED ISSUE #149: Default `configPath` for addons points to dummy app's config path.
As per #133, we now allow a custom `configPath` to be set in an engine's `package.json`. If `ember-addon/configPath` is set, then this value will be used instead of the default path of `config`.

Unfortunately, `ember-addon/configPath` _is_ set by default in new addon projects, and the default value is `"tests/dummy/config"`. New engines are now using this the environment settings from the dummy app (`tests/dummy/config/environment.js`) instead of `config/environment.js`.

@rwjblue any idea how to best untangle this while still allowing, but not requiring, custom config paths?

dgeb:
Instead of overloading the usage of `configPath`, engines need their
own unique setting to specify their configuration file path.

[Fixes #149]

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -45,8 +45,8 @@ module.exports = {
     options.engineConfig = function(env, baseConfig) {
       var configPath = 'config';
 
-      if (this.pkg['ember-addon'] && this.pkg['ember-addon']['configPath']) {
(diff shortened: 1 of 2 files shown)


dgeb:
cc @cklimkowsky @trentmwillis

PR #160, opened 2016-07-15
Document routing concerns for lazy engines

trentmwillis:
- Document route-serializer usage in ember-engines and how it differs from behavior outside of engines.
- Document change in loading behavior for failed transitions when crossing an engine boundary.


dgeb:
Thanks @trentmwillis!

PR #166, opened 2016-07-22
[0.2] Fix: Query parameters are not rehydrated

### LINKED ISSUE #163: Query params do not work
I have the following route in my engine:

```js
// index/controller.js
export default Controller.extend({

  queryParams: ['sort', 'sortOrder'],

  sort: null,
  sortOrder: null,

  actions: {
    updateSort(sort) {
      this.set('sort', sort);
    },
    updateSortOrder(sortOrder) {
      this.set('sortOrder', sortOrder);
    }
  }
});
```

```hbs
{{! index/template.hbs }}
<button {{action 'updateSort' 'name'}}>sort = name</button>
<button {{action 'updateSort' 'idTag'}}>sort = idTag</button>
<button {{action 'updateSortOrder' 'DESC'}}>sortOrder = DESC</button>
<button {{action 'updateSortOrder' 'ASC'}}>sortOrder = ASC</button>
```

However, apparently the query params overwrite each other somehow.

1.) Initially, no query param is there.
2.) I press the first button, sort=name
3.) I press the second button, sort=idTag
4.) I press the third button, sortOrder=DESC, **sort is gone!**
5.) I press the first button again, sort=name, **sortOrder is gone!**

The exact same template/controller works fine in the host app.

dnachev:
The problem is the cache, which is used to keep track of the query param values. This cache is not shared between the application and the engine, but the routing is entirely done in the application. This causes the hydrating of unspecified query parameters during transitioning to fail, because the query param value is actually stored in the engine cache, but the hydration is done from the application cache.

My first PR for this addon, so any feedback is welcome. If you think this is correct fix I can try to add an acceptance test as well.

### DIFF
--- a/addon/-private/engine-instance-ext.js
+++ b/addon/-private/engine-instance-ext.js
@@ -219,7 +219,6 @@ EngineInstance.reopen({
       'view:toplevel',
       'route:basic',
       'event_dispatcher:main',
-      P`-bucket-cache:main`,
       'service:-routing'
     ].forEach((key) => {
       this.register(key, parent.resolveRegistration(key));
@@ -228,6 +227,7 @@ EngineInstance.reopen({
     [
       'renderer:-dom',
       'router:main',
+      P`-bucket-cache:main`,
       '-view-registry:main'
     ].forEach((key) => {
       this.register(key, parent.lookup(key), { instantiate: false });
--- a/tests/acceptance/routeable-engine-demo-test.js
+++ b/tests/acceptance/routeable-engine-demo-test.js
@@ -30,6 +30,20 @@ test('can deserialize a route\'s params', function(assert) {
   });
 });
 
+test('can restore unspecified query parameters', function(assert) {
+  assert.expect(2);
+
+  visit('/routable-engine-demo/blog/post/1?lang=English');
+  click('.routable-post-comments-link');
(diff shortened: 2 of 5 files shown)


trentmwillis:
@dnachev unsure about correctness (I'll defer to @dgeb), but it would be good to go ahead and add a previously failing test to prevent regressions.

dnachev:
@trentmwillis I've added a test, although I'm not sure it recreates the problem in issue #163. However, it fails and my change fixes the test, so at least this case should be handled. I wouldn't be surprised if it fixes the query params reset from the issue as well, but I have no guarantees for it.

dgeb:
This looks like the right fix to me.

@rwjblue - r?

rwjblue:
Yes, this seems correct to me.  The bucket cache should be shared.

@dnachev - Can you please submit this patch to Ember also?  Somewhere in here: https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/engine-instance.js#L180-L183

dgeb:
Published in v0.2.8

dgeb:
@dnachev I added https://github.com/dgeb/ember-engines/issues/168 to track the push upstream. Please comment there if you plan to take this on. Thanks!

Issue #164, opened 2016-07-20
Guard Engine config for SSR


trentmwillis:
With our SSR setup we're running into an issue where we error out when trying to read in an Engine's config due to attempting to use jQuery (https://github.com/dgeb/ember-engines/blob/master/lib/engine-config-from-meta.js#L7).

We have a hack in place for the app's `config/environment` but that isn't possible for the Engines. We would still like access to the config, but we're unsure of the best way around the DOM access. Any idea's how FastBoot handles this for the standard config?

cc/ @thousand @kratiahuja

rwjblue:
Fastboot overrides the `app-name/config/environment` module to avoid the jQuery usage ([here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/187)).

kratiahuja:
Fastboot is doing the same way as we do with a little difference of providing an API as well. We probably need to unify our work with fastboot. 

For engines, we probably need a hook to do this kind of replacement.

trentmwillis:
Okay, sounds like we want https://github.com/ember-cli/ember-cli/pull/5582 to land so that all the modules can flow through the same hook.

trentmwillis:
@rwjblue what would you think about doing [something similar to this](https://github.com/null-null-null/ember-get-config/blob/master/index.js#L15-L19)? Where we just generate an actual module with the JSON stringified in it? I don't see a need to have the Engine's configuration inlined in the DOM, but I could be missing something.

rwjblue:
Embedding the config into the asset means you cannot promote an asset from one environment to another.

trentmwillis:
> Embedding the config into the asset means you cannot promote an asset from one environment to another.

@rwjblue can you explain this a bit more? Or perhaps provide an example of where it'd be bad? I read it like a dozen times but am still unsure what you mean.

trentmwillis:
#175 was enough to allow us to work around the blocking issue. There likely needs to be a standard way of adding configs, but I don't think that discussion should be had in this repo, thus I am closing this issue.

PR #175, opened 2016-08-05
Generate engine config in separate method

trentmwillis:
An attempt at addressing #164. This provides a hook which can be overriden to return a version of the config module that works in non-browser environments.

I'm open to other options, but this is the best one I could come up with currently.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -110,6 +110,20 @@ module.exports = {
     };
 
     /**
+      Returns the contents of the module to be used for accessing the Engine's
+      config.
+
+      @public
+      @method getEngineConfigContents
+      @return {String}
+    */
+    options.getEngineConfigContents = options.getEngineConfigContents || function() {
+      var configTemplatePath = path.join(__dirname, '/engine-config-from-meta.js');
+      var configTemplate = fs.readFileSync(configTemplatePath, { encoding: 'utf8' });
+      return configTemplate.replace('{{MODULE_PREFIX}}', options.name);
+    };
+
+    /**
       Returns a given type of tree (if present), merged with the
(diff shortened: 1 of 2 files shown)


nathanhammond:
As this isn't targeting master I'm okay with landing it. Are we going to need to do this again for the master branch?

trentmwillis:
@nathanhammond either this or something else. I think there is a better solution out there, but would require much more work (essentially we should have one hook for defining configs and it should do the right thing for different environments).

nathanhammond:
I feel like this is a mess we [keep just throwing flags at](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1698-L1722).

This is more graceful in that it exposes a hook, but an additional hook is more API that we have to maintain. I think I'd like `config` in general to be resolved to JSON and then have something in the pipeline that gives you options on what you want to do with JSON files: convert to JS modules, convert to meta tags, leave as is...

trentmwillis:
> I think I'd like config in general to be resolved to JSON and then have something in the pipeline that gives you options on what you want to do with JSON files: convert to JS modules, convert to meta tags, leave as is...

More-or-less what I was thinking in the above statement. We just need a unified way of dealing with these things. One (sort of close) approach could be the multi-meta modules idea.

trentmwillis:
@dgeb or @rwjblue thoughts on landing this? It's blocking us from upgrading to a newer Ember

rwjblue:
Changes seem good. Should we target master in a separate PR?

trentmwillis:
> Should we target master in a separate PR?

Seems like we should. Will open that PR.

dgeb:
Ok, released in v0.2.11. Thanks again @trentmwillis!

PR #178, opened 2016-08-09
Ember 2.7 compatibility

trentmwillis:
Attempting to get 2.7 compatibility. Would be nice to be able to upgrade before the flag is enabled in 2.8.

### DIFF
--- a/addon/-private/engine-ext.js
+++ b/addon/-private/engine-ext.js
@@ -10,7 +10,6 @@ const P = emberRequire('container/registry', 'privatize');
 const topLevelViewTemplate = emberRequire('ember-htmlbars/templates/top-level-view');
 
 const {
-  OutletView,
   TextField,
   TextArea,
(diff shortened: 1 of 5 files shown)


addon/-private/engine-ext.js
```diff
@@ -52,7 +51,6 @@ Engine.reopen({
 
       registry.register('template:-outlet', topLevelViewTemplate);
       registry.injection('route', '_topLevelViewTemplate', 'template:-outlet');
-      registry.register('view:-outlet', OutletView);
```
nathanhammond: I'm unsure why you can drop this?

trentmwillis: Looking through the Ember 2.6 code, seems like it is redundant, as it gets registered via the super of `buildRegistry`.

dgeb:
Thanks @trentmwillis!

PR #183, opened 2016-08-16
Add `lazyLoading` property to README

SaladFork:
Since not including this property has been deprecated.

We should also consider improving the blueprints to include this property out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,17 @@ export default Eng;
 It's important that `modulePrefix` be set in `config/environment.js` so that
 it can be referenced in `addon/engine.js`.
 
+### Lazy Loading Engines
+
+You must also declare in your Engine's `index.js` file whether or not the engine should be lazy loaded. Until lazy loading is supported, this should be set to `false`:
+```js
+var EngineAddon = require('ember-engines/lib/engine-addon');
+module.exports = EngineAddon.extend({
+  name: 'ember-blog',
+  lazyLoading: false
+});
+```
+
 ### Routable Engines
 
 Routable engines should declare their route map in a `routes.js` file.


trentmwillis:
Maybe it's just me, but I feel like we should turn off the deprecation until support for lazy loading has actually landed.

rwjblue:
The idea was to get folks specifying `false` because anything that works today is likely to need to be `lazyLoading: false`.  When we add support for lazy loading in https://github.com/dgeb/ember-engines/pull/173 folks can just update, and start fiddling with changing the value...

SaladFork:
I think adding it to the blueprints would help a lot. It was less-than-nice to create a new engine and have a deprecation warning already.

nathanhammond:
My comments in Slack:

Because we want to make the user explicitly specify their intention. It's easy to accidentally troll yourself with this and we wanted to make that require input on the user. We also want to consider "lazy-by-default" which means that we need to get all unspecified engines out of existence.

Basically if we force you to specify now we get a choice of the default behavior later. Otherwise we have to choose the default now, and I'm not sure we're ready to make that call.

mike-north:
Discussion and deliberation of future defaults of engines notwithstanding, the current state of things is a bit confusing for all but the deepest experts.

Biggest pain point I've seen w/ early adopters: Without any additional information, there's nothing to indicate that `lazyLoading: true` is not going to get you anywhere (yet). They end up trying to run down that path, only to find that the build tool changes that lazy loading needs haven't yet landed.

trentmwillis:
@rwjblue / @nathanhammond / @dgeb I'm in favor of landing this so that we have some documentation around it and can point developers to it when they ask questions. It does note that lazy loading is not quite ready, so I think that is also beneficial.

knownasilya:
Should the parent app be able to override whether an engine it uses is loaded lazily?

dgeb:
@knownasilya yes, that is the plan. The parent app should have final say in the matter.

PR #173, opened 2016-07-28
[WIP] Lazy load build

nathanhammond:
## TODO
- [ ] Include child addons' addon trees which are not in the host application inside of the engine bundle.
- [ ] `app.import` is probably having a bad day.

PR #184, opened 2016-08-17
Lazy Load Runtime

trentmwillis:
This is the complement to the Lazy Load Build PR (#173). There are three main parts to this PR:

1. Generate an Asset Manifest from the build assets of lazy engines (accomplished via [ember-asset-loader](https://github.com/trentmwillis/ember-asset-loader)).
2. Load lazy Engine assets dynamically at runtime (this is the bulk of the PR)
3. Modify default query params behavior such that they are ignored for Engines (otherwise we'll blow up on lazy Engines).

Todo:
- [x] Fix `handler.actions is undefined` error [[PR](https://github.com/emberjs/ember.js/pull/14087)]
- [x] Fix tests failing due to lazy loading
  - [x] Determine approach for testing lazily loaded code [`preloadAssets`]
  - [ ] Move `preloadAssets` to `ember-asset-loader` [[PR](https://github.com/trentmwillis/ember-asset-loader/pull/15)]
- [x] Write tests to verify behavior
- [x] Cleanup / document / polish

### DIFF
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -1,0 +1,203 @@
+import Ember from 'ember';
+import emberRequire from './ext-require';
+
+const hasDefaultSerialize = emberRequire('ember-routing/system/route', 'hasDefaultSerialize');
+const EmptyObject = emberRequire('ember-metal/empty_object');
+
+const {
+  Logger: {
+    info
+  },
+  Router,
+  RSVP,
+  assert,
+  get,
+  getOwner
+} = Ember;
+
+Router.reopen({
+  assetLoader: Ember.inject.service(),
+
+  /**
+   * We skip doing default query params stuff when attempting to go to an
+   * Engine route.
+   *
+   * @override
+   */
+  _prepareQueryParams(routeName) {
+    if (this._engineInfoByRoute[routeName]) {
+      return;
+    }
+
+    return this._super(...arguments);
+  },
+
+  /**
+   * We override this to fetch assets when crossing into a lazy Engine for the
+   * first time. For other cases we do the normal thing.
+   *
+   * @override
+   */
+  _getHandlerFunction() {
+    let seen = new EmptyObject();
+    let owner = getOwner(this);
+
+    return (name) => {
+      let engineInfo = this._engineInfoByRoute[name];
+
+      if (engineInfo) {
+        return this._getEngineInstance(engineInfo).then((instance) => {
+          let handler = this._internalGetHandler(seen, name, engineInfo.localFullName, instance);
+
+          if (!hasDefaultSerialize(handler)) {
+            throw new Error('Defining a custom serialize method on an Engine route is not supported.');
+          }
+
+          return handler;
+        });
+      }
+
+      // If we don't cross into an Engine, then the routeName and localRouteName
+      // are the same.
+      return this._internalGetHandler(seen, name, name, owner);
+    };
+  },
+
+  /**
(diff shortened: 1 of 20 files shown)


addon/-private/router-ext.js
```diff
@@ -0,0 +1,131 @@
...
+
+    return this._super(...arguments);
+  },
+
+  _getHandlerFunction() {
+    let seen = {}; // Originally EmptyObject
+    let owner = getOwner(this);
+
+    return (name) => {
+      console.log('getting', name);
```
stefanpenner: for dev?

addon/-private/router-ext.js
```diff
@@ -0,0 +1,131 @@
...
+
+      let engineInfo = this._engineInfoByRoute[name];
+
+      if (engineInfo) {
+        return this._getEngineInstance(engineInfo).then((instance) => {
+          if (instance) {
+            let handler = __getHandler(engineInfo.localFullName, instance);
+
+            // if (engineInfo && !hasDefaultSerialize(handler)) {
+            //   throw new Error('Defining a custom serialize method on an Engine route is not supported.');
```
MiguelMadero: Any reason this is commented out? I think it makes sense to throw.

tests/helpers/preload-assets.js
```diff
@@ -0,0 +1,26 @@
+import Ember from 'ember';
+import AssetLoader from 'ember-asset-loader/services/asset-loader';
```
MiguelMadero: It seems a bit odd to import the service and create an instance instead of injecting it and using the same instance. 

I'm worried this could bite us later if someone is explicitly preloading assets, but we're using a different [cache](https://github.com/trentmwillis/ember-asset-loader/blob/master/addon/services/asset-loader.js#L176)

addon/-private/router-ext.js
trentmwillis: Work in progress still.

tests/helpers/preload-assets.js
trentmwillis: This is a test helper. Needs to be refined a bit, still a work in progress.

rwjblue: We should use `addon-test-support` for any publicly exposed test helpers (not sure if this one is meant for just us in the addon or for all users).

trentmwillis: I actually think this will move to the `ember-asset-loader` repo. But duly noted, will probably use `addon-test-support`.

MiguelMadero:
LGTM

addon/-private/router-ext.js
```diff
@@ -0,0 +1,202 @@
...
+  },
+
+  /**
+   * We override this to fetch assets when crossing into an Engine. For other
+   * cases we do the normal thing.
+   *
+   * @override
+   */
+  _getHandlerFunction() {
+    let seen = Object.create(null);
```
chadhietala: I'm not sure if we want to use an object here. I believe this object is going to be prone to a lot of cache misses and would also be susceptible to blowing up the IC at site where we check the cache. I may be wrong so @krisselden should probably comment sense he's been working in this domain.

trentmwillis: This is an approximation of what is happening upstream ([ref](https://github.com/emberjs/ember.js/blob/c94d277b82e57962953ccf601b8862016cb46f29/packages/ember-routing/lib/system/router.js#L563)). Will have it actually using `EmptyObject` in the next update.

dgeb:
Thanks @trentmwillis!

PR #185, opened 2016-08-25
Use preloadAssets helper from ember-asset-loader

trentmwillis:
Follow up to https://github.com/dgeb/ember-engines/pull/184.

`ember-asset-loader` is a more logical place for the `preloadAssets` helper to come from (since it is not specific `ember-engines`).

Also worth noting that the preloading assets approach was chosen to avoid introducing order-dependency into test suites and to make it easy to do unit and integration tests (which will fail without first loading assets). It is also safe to preload assets since the actual act of loading them should have test coverage through this addon and the `ember-asset-loader` addon.


nathanhammond:
LGTM

Issue #10, opened 2015-12-22
Update ember-resolver addon to use `resolveEngine`.


rwjblue:
For now, apps that consumes engines must use the `ember-engines/resolver` module instead of `ember/resolver` or `ember-resolver`. Once we are confident in the `resolveEngine` implementation added in #9 we should push those changes upstream into ember-resolver (possibly after publishing a beta version of this addon).  Then we can remove `addon/resolver.js`.

asakusuma:
@rwjblue are we confident in `resolveEngine` now? If so I will take a crack at this.

rwjblue:
:shipi

PR #187, opened 2016-09-01
Upstream resolver changes

trentmwillis:
Dependent on: https://github.com/ember-cli/ember-resolver/pull/151.

Also, starts testing against the Beta branch of Ember. Figured we should probably do that if 2.8 is supposed to be compatible with Engines.

Closes: #10


trentmwillis:
@dgeb and @rwjblue this is passing in Beta, but Canary is failing due to a build issue. Dug into it a bit, but couldn't figure it out. Seems potentially similar to this old issue: https://github.com/emberjs/ember.js/issues/13347

trentmwillis:
Failure seems to have resolved itself. I think this should be good to merge.

chadhietala:
LGTM

dgeb:
Thanks @trentmwillis :shipit:

Issue #159, opened 2016-07-15
[0.2] Controller.transitionToRoute() doesn't work for engine routes


dnachev:
The problem is caused, because the `Controller.transitionToRoute()` relies on the `router.transitionTo()`, which is not patched to be engine aware.

trentmwillis:
@dnachev any chance you could submit a fix for this? Seems like it should be straightforward.

trentmwillis:
I've opened a PR to address this upstream: https://github.com/emberjs/ember.js/pull/14176

dnachev:
Opened PR for v0.2 branch: https://github.com/dgeb/ember-engines/pull/188

trentmwillis:
Closed via #188.

PR #188, opened 2016-09-01
Make transitionToRoute() and intermediateTransitionTo() work in engines

### LINKED ISSUE #186: [0.2] Route.intermediateTransitionTo() doesn't prefix the routes
It needs to have the same prefixing done as the `transitionTo`.

dnachev:
The PR fixes #186 and #159 in v0.2 branch.

I've added tests, but the one for `intermediateTransitionTo()` is somewhat clunky. I couldn't think of a better way.

### DIFF
--- a/addon/-private/utils.js
+++ b/addon/-private/utils.js
@@ -1,0 +1,32 @@
+import Ember from 'ember';
+
+const {
+  getOwner,
+  Error: EmberError
+} = Ember;
+
+/*
+  Returns an arguments array where the route name arg is prefixed based on the mount point
+*/
+export function prefixRouteNameArg(funcName, ...args) {
+  let routeName = args[0];
+  let owner = getOwner(this);
+  let prefix = owner.mountPoint;
+
+  // only alter the routeName if it's actually referencing a route.
+  if (owner.routable && typeof routeName === 'string') {
+    if (resemblesURL(routeName)) {
+      throw new EmberError(`${funcName} cannot be used for URLs. Please use the route name instead.`);
+    } else {
+      routeName = `${prefix}.${routeName}`;
+      args[0] = routeName;
+    }
+  }
+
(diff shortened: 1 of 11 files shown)


trentmwillis:
@dgeb / @rwjblue thoughts on merging this and doing one final `v0.2` release? We'll likely want to port these tests forward as well (this behavior exists upstream now).

dgeb:
@trentmwillis I'm in favor. Do you have time time to review that this PR matches the upstream changes?

addon/-private/utils.js
```diff
@@ -0,0 +1,32 @@
...
+*/
+export function prefixRouteNameArg(funcName, ...args) {
+  let routeName = args[0];
+  let owner = getOwner(this);
+  let prefix = owner.mountPoint;
+
+  // only alter the routeName if it's actually referencing a route.
+  if (owner.routable && typeof routeName === 'string') {
+    if (resemblesURL(routeName)) {
+      throw new EmberError(`${funcName} cannot be used for URLs. Please use the route name instead.`);
```
trentmwillis: Forgot to make this dynamic upstream...will have to do that soon.

trentmwillis:
@dgeb here's the upstream PR for reference: https://github.com/emberjs/ember.js/pull/14262.

A quick look over shows that `Controller#replaceRoute` is missing in this PR, but that is private API so I'm unsure if we need to support it here. Otherwise, it looks solid to me.

dgeb:
@trentmwillis Thanks for the review. And good catch about needing to make `prefixRouteNameArg` dynamic upstream.

@dnachev Could you please update `Controller#replaceRoute` to match `Controller#transitionToRoute` (just like upstream)?

dgeb:
Merged via #207. Release in v0.2.12. Thanks @dnachev!

PR #191, opened 2016-09-09
ApplicationInstance.boot() fails because of missing engine parent

dnachev:
There is a check for set engine parent, when trying to boot an `ApplicationInstance`, created with `Application.buildInstance()`.

The test pass on Ember 2.6.2 and fails on Ember 2.7.3. The reason is that in 2.6, the ApplicationInstance overrode both `boot` and `_bootSync`, with its own copy and the check for the engine parent is not done. In 2.7, `ApplicationInstance` only overrides `_bootSync`, which leaves the engine parent check to be done in the `boot()` section.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "ember-engines",
   "dependencies": {
-    "ember": "~2.7.0",
+    "ember": "~2.7.3",
     "ember-cli-shims": "0.1.1",
     "ember-cli-test-loader": "0.2.2",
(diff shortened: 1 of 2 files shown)


trentmwillis:
Looking into this. Seems like it could be an issue when using the `visit` API.

kratiahuja:
@trentwillis if it is a issue with 'visit' API it should fail on fastboot.

Cc: @tomdale

dgeb:
Fixed via #192

PR #198, opened 2016-09-13
Remove usage of EmptyObject.

### LINKED ISSUE #196: [master] - ember-metal/empty_object is not found (at first)
**The failure**
in router-ext.js, [an empty object constructor is imported from ember](https://github.com/dgeb/ember-engines/blob/master/addon/-private/router-ext.js#L5). The first time this is required using ember canary (2.10.0-canary+f66b62e2), the loader doesn't find it, and we fall back to `missingModule`. Things generally fall apart from there.

**After things have failed**
[Using the same technique to load ember internal modules](https://github.com/dgeb/ember-engines/blob/master/addon/-private/ext-require.js#L4), once the attempted (and failed) app booting completes actually does find the module in question
screen shot 2016-09-13 at 2 00 38 pm

FWIW making the following change to router-ext.js at least gets the app loading.
```js
const EmptyObject = Object; //emberRequire('ember-metal/empty_object');
```

Ember engines version: `fdddb1bda5999ab77d574099ce00f1b448d43721` (master)
Ember version: `2.10.0-canary+f66b62e2`

rwjblue:
`EmptyObject` in Ember is an attempt to make creating `Object.create(null)` objects faster by avoiding the creation of an object with a null prototype.

In our case, these objects are not being created in a hot path repeatedly so the performance concern is not valid. Using `Object.create(null)` directly reduces our coupling to changes in Ember and embraces "the platform" (:troll:).

Fixes #196.

### DIFF
--- a/addon/-private/link-to-component.js
+++ b/addon/-private/link-to-component.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { attributeMungingMethod } from './link-to-utils';
 
 const {
   LinkComponent,
@@ -8,7 +9,7 @@ const {
 } = Ember;
 
 export default LinkComponent.extend({
-  willRender() {
+  [attributeMungingMethod]() {
     this._super(...arguments);
 
     let owner = getOwner(this);
--- a/addon/-private/link-to-external-component.js
+++ b/addon/-private/link-to-external-component.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import { attributeMungingMethod } from './link-to-utils';
 
 const {
   LinkComponent,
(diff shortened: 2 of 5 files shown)


rwjblue:
CI failures are partially due to https://github.com/trentmwillis/ember-asset-loader/pull/20.

rwjblue:
OK, is good for review.  CI will be fixed once https://github.com/trentmwillis/ember-asset-loader/pull/20 lands and is released (I've tested with that commit locally and tests pass).

trentmwillis:
Looks good to me and Travis green :+1:

krisselden:
@rwjblue also chrome fixed the bug that made it slow

PR #200, opened 2016-09-14
[WIP] Simplify engine exports.

rwjblue:
- Avoid relying on initializer to setup `.reopen`'s needed.
- Use `.extend()` when possible (instead of `Engine.reopen` we can
  `.extend()`)
- Remove initializer (it is not needed internally).
- Remove all but one usage of `Ember.__loader.require` (remaining usage
  is for `hasDefaultSerializer`).


tests/unit/initializers/engines-test.js
```diff
@@ -15,8 +14,6 @@ module('Unit | Initializer | engines', {
 
 // Replace this with your real tests.
 test('it works', function(assert) {
-  EnginesInitializer.initialize(application);
-
   // you would normally confirm the results of the initializer here
   assert.ok(true);
 });
```
nickiaconis: This test file doesn't do anything productive anymore. Do you have plans for it or should it be removed?

rwjblue: good catch, will remove

trentmwillis:
I tested this out locally, and it breaks the lazy loading runtime behavior, since the Router extension doesn't occur until after the Router instance has been created.

I'm not entirely sure why the tests don't break, probably has to do with preloading the assets.

scalvert:
@rwjblue any thoughts on how to resolve this?

rwjblue:
@scalvert - Yes, just lack of time to complete.  Ping me in slack if you would like to pick it up.

scalvert:
Will do

PR #211, opened 2016-09-23
[WIP] Keep everything in the styles tree originally present.

nathanhammond:
I'm not sure that this is correct. I need to research where these assets get moved to within the tree in [`EmberApp`](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js)


rwjblue:
Closing, this is pretty stale...

PR #224, opened 2016-10-14
Add missing semicolon.

abuiles:
This seems to be intentional? My linter complains about it ;(


dgeb:
@abuiles This was not intentional AFAIK, so thanks for correcting!

PR #225, opened 2016-10-17
Add gitignore file for blueprints

knownasilya:
Npm installing within these dirs asks to commit the new directories.


trentmwillis:
In-repo-engines (and addons) should use the host repo's `.gitignore` file. Additionally, you shouldn't need to `npm install` or `bower install` from within an in-repo-engine. Closing, for now, but if there is a use case I have missed, then we can revisit, but I'm fairly confident this should be solved in the host repo.

Issue #209, opened 2016-09-23
Improve query params behavior


trentmwillis:
With lazy Engines coming and the router now returning a Promise from `getHandler`, we need to add more test cases for query parameter behavior. Our current behavior was a naive stubbing of `_prepareQueryParams` which seems to fail in many scenario more complex than simple transitions or direct navigation. Trying to integrate `master` into our app currently has been difficult due to this reason.

TODO:

* [x] Test case for QP-only link-to transition
* [x] Test case for QP-only programmatic transition
* [x] Test case for redirect transition with QP
* [x] Test case for redirect transition after model with QP

Feel free to chime in with other potential test cases.

trentmwillis:
Fixed via #227.

PR #227, opened 2016-10-17
Fix query params and add tests

trentmwillis:
Addresses #209.

This should cover a large swath of Engines + QP use cases. There is a [super tiny change that needs to happen upstream](https://github.com/emberjs/ember.js/pull/14487) in order for this to pass.

### DIFF
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -18,14 +18,16 @@ Router.reopen({
   assetLoader: Ember.inject.service(),
 
   /**
-   * We skip doing default query params stuff when attempting to go to an
-   * Engine route.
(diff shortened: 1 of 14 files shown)


trentmwillis:
Tests are now passing in Canary.

rwjblue:
Restarted beta build (2.10.0-beta.1 was cut last night)...

knownasilya:
This deserves a new release 👍

nathanhammond:
@knownasilya Only updated against `master`, would require backporting to the 0.3.X branch or waiting a bit until we're comfortable with our build stuff (testing that in our app now-ish, at which point we will know better).

PR #228, opened 2016-10-18
Load config based on `storeConfigInMeta` option

kamalaknn:
I've tried to feature pair `storeConfigInMeta` option from ember-cli. Ported from [ember-cli's implementation](https://github.com/ember-cli/ember-cli/blob/f863535532ab5c5884bab77df8f3a467ab60f702/lib/broccoli/ember-app.js#L1754-L1761).

I am unsure about `this.app` for nested engines though. Feedback welcome 😄


nathanhammond:
@trentmwillis and I talked about something like this today for one of our own use cases. There is a weird interplay of issues here between Fastboot, lazy engines, app code, and our current asset manifest insertion strategy (which consistently rewrites the meta tag as time goes on since we never know when we're "done").

This is something that needs further design discussion prior to us deciding if and how to land a feature that does this. Regardless, this is a fantastic straw man.

Would you be willing to rebase onto `master` and get tests passing?

kamalaknn:
@nathanhammond Sure. I will give it a go 👍

kamalaknn:
Rebased moved against master. Travis is green now 😃

joebartels:
Could someone share status on this story. I still have this issue. Is there a known workaround?

Using 
Ember-Engines v0.5.0
Ember  v2.12.0
Ember-Cli-Fastboot 1.0.0-beta.15

krisselden:
We don't want to inline the config into the JS.  ember-fastboot needs a way an addon can add to the sandbox so that engines can get the config from there.

Also, it should be ok for an engine not to have any configuration.

kamalaknn:
@rwjblue I could see that [`engine-config-from-meta.js`](https://github.com/ember-engines/ember-engines/blob/master/lib/engine-config-from-meta.js) is already on the project. I could be wrong as well, it's been quite some time since I worked with ember/ember-cli. I might have lost context.

skaterdav85:
Does anyone have a workaround for `storeConfigInMeta: false`?

ASH-Michael:
@rwjblue It looks like this PR is just about at the finish line after the conflicts are resolved. Is there anything else that this PR still needs?

chendraayan:
Just wondering what happened to this story. Does anyone have a workaround for `storeConfigInMeta: false`?

villander:
@rwjblue do we still need this? what is the relation of storeConfigInMeta with ember engines and fastboot?

puopg:
Id love to be able to read the config in without the meta tags

villander:
@stefanpenner I think that we need of a enhancement to read the config without meta tag, what about?

puopg:
Given how old this PR is, and since I cant update this one, i opened #633

PR #234, opened 2016-10-26
Document how to "share" things between engine and parent app.

abuiles:
Making the answer from https://github.com/dgeb/ember-engines/issues/139 more visible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,36 @@ export default Eng;
 Currently, only services and route paths (see below) can be shared across the
 parent/engine boundary.
 
+### Sharing Components and More
+
+If you want to share components, helpers or utils between a parent and
+an engine, then create an addon and add it as a dependency of both
+your application and engine.
+
+```json
+{
+  "dependencies": {
+    "common-components": "1.0.0"
+  }
+}
+```
+
+In-repo addons can be consumed from an engine too. To do so, add it to the
+engine's `package.json`:
+
+```
+// lib/my-in-repo-engine/package.json
+  "ember-addon": {
+    "paths": [
+      "../shared-components"
+    ]
+  }
+```
+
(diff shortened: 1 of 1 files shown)


abuiles:
I spent a couple of minutes trying to figure this out. Documenting it here for people who is adopting engines too.

nathanhammond:
Thank you for jumping in here and adding some documentation! One minor nitpick and let's land it. :smile:

README.md
```diff
@@ -213,6 +213,22 @@ export default Eng;
...
+your application and engine.
+
+In-repo addons can be consumed from an engine. To do so, add it to the
+engine's `package.json`:
+
+```
+// lib/my-in-repo-engine/package.json
+  "ember-addon": {
+    "paths": [
+      "../shared-components"
```
abuiles: @nathanhammond makes sense, the point I want to make is that is still possible to consume in-repo addons, if you are using in-repo engines, maybe I should make that clear?

nathanhammond: Yes, let's split that out as well into a separate section instead of trying to teach two concepts with one code example. That's also worth teaching, but I want to couch it in language that demonstrates how it is approximately breaking isolation.

abuiles: @nathanhammond updated. Added as a section with both scenarios.

I noticed there is some duplication between the README and articles under guides. Maybe we can publish stuff under guides as a github page and leave the README for dev notes only? cc @knownasilya

knownasilya: I can submit a PR, if that's good 👍  like https://blog.abuiles.com/ember-watson/ 😄

abuiles:
@nathanhammond ping

MelSumner:
@abuiles any idea why the AppVeyor build failed?

abuiles:
@MelSumner no, also this is pretty old, so it might not be relevant any more.

Issue #244, opened 2016-10-27
`intermediateTransitionTo` once inside an Engine


trentmwillis:
An edge case we hadn't considered has cropped up.

So imagine we're already in an Engine. We know that any `intermediateTransitionTo` needs to be synchronous, but we also know that any route in an Engine is always loaded asynchronously (even if it is just `Promise.resolve()`).  Thus, if you’re in an Engine and do an `intermediateTransitionTo` (such as to a `loading` state/route) then you're gonna have a bad time.

One way to fix this would be to return handlers synchronously, once you are in an Engine. This can be accomplished by setting a flag on the `engineInfo` object associated with each Engine route.

Alternatively, we can exempt `loading`/`error` routes from the lazy loading semantics, but this seems brittle.

I'm open to other suggestions, but this is what I have so far.

cc @nathanhammond @rwjblue

trentmwillis:
Fixed via #246.

PR #246, opened 2016-10-28
Return handlers synchronously from previously visited Engines

trentmwillis:
Addressing #244 with one of the suggested ideas:

> return handlers synchronously, once you are in an Engine.

---

The first approach at the above idea was to return handlers synchronously for any Engine that has its assets loaded. This approach was troublesome largely due to testing. If we return handlers synchronously when the Engine's assets are loaded and we preload all assets in our testing environment, then we will never hit the asynchronous case in our testing, which does not reflect real world usage.

Additionally, this could potentially cause issues when dealing with nested Engines or Engines mounted in more than one location.

---

The second approach, and current one, is to return handlers synchronously if the Engine instance already exists. This means the first load for an Engine will always be async, even in testing and for eager Engines, and subsequent visits will always be synchronous. This even works when you use nested Engines or Engines mounted in multiple locations.

This also aligns closely with the intent of the original idea, that once you are _in_ an Engine, handlers are resolved synchronously. To be _in_ an Engine, it must be loaded and have an instance.

---

The divergence in behavior between first and second visits is a bit worrisome to me, but I think this is likely to be the most robust solution. I would eventually like us to get to a place where all handlers are async always, so we can stop this dual path stuff, but I don't see a way to make that work with `intermediateTransitionTo`. Will likely be something to keep in mind for Ember 3.0.


trentmwillis:
This is ready to go. AppVeyor is just being dumb.

trentmwillis:
@rwjblue added another assertion to the test; it was failing previously which is why I forgot to check that it actually called `intermediateTransitionTo`.

rwjblue:
Thanks!

PR #249, opened 2016-11-01
Don't use rollup-funnel for non-lazy builds

trentmwillis:
Improves build and re-build performance.

cc @nathanhammond


nathanhammond:
Looking at this now I'm unsure why this code path exists. We may find out later, but it seems safe to remove to me. :shipit:

trentmwillis:
Hoping it was just an oversight from being too close to the problem when it was originally written. Should we test against your test app?

nathanhammond:
I'm in the process of doing that to confirm all things look good.

nathanhammond:
(Reasonable order of operations: who needs it?)

PR #250, opened 2016-11-02
Work smarter, not harder.

nathanhammond:
Currently `ember-engines` does unnecessary work in an all-eager scenario to remove and re-add routes. This stops that madness.

Manually tested that this produces identical output for the [`lazy-engine-testing`](https://github.com/nathanhammond/lazy-engine-testing) repo.


trentmwillis:
Verified this works in our application at LI, so merging...

PR #252, opened 2016-11-08
Update so that addons can accept asset objects w/ environment properties

ima007:
Context:
We have an addon importing a vendor file and specifying both a development and production version of the file. I received an error when trying to use the environment object version of the asset argument, which I traced down to the `asset.replace` (which assumes a string being passed in).

In our case, we were able to work around this simply by using a string asset path -- but thought this might be a valid use case to roll up to ember-engines.

Note:
Haven't tested this, yet. Opening mainly for comments in case there's another recommended approach or reason for sticking with string paths.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -299,6 +299,7 @@ module.exports = {
         var host = originalFindHost.call(this);
         var target = this._findHost();
 
+        target.env = host.env;
         target._import = host._import;
         target._getAssetPath = host._getAssetPath;
         target.otherAssets = host.otherAssets;
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -294,6 +294,7 @@ module.exports = {
         return buildExternalTree.call(this);
       };
 
+
       this.import = function(asset, options) {
         var host = originalFindHost.call(this);
         var target = this._findHost();
@@ -305,6 +306,9 @@ module.exports = {
 
         // We're delegating to the upstream EmberApp behavior for eager engines.
         if (this.lazyLoading !== true) {
+          // Asset can be an object with environment properties.
+          asset = target._getAssetPath(asset);
+
           // This is hard-coded in Ember CLI, not tied to treePaths.
           asset.replace(/^vendor/, '');
         }


trentmwillis:
This approach seems reasonable to me. Would just need to add an example use case in [engine-testing](https://github.com/ember-engines/engine-testing).

rwjblue:
The upstream `_getAssetPath` implementation uses `this.env` to determine which environment to actually use. Since `this.env` is `undefined` for addons, I _believe_ that this PR avoids the error but does not actually result in importing the asset. 

The solution would be to add an `env: function() { return process.env.EMBER_ENV; }` function to engines in `lib/engine-addon.js`.

buschtoens:
@rwjblue Are you sure you're not confusing the context for `this`?

In [`_getAssetPath`](https://github.com/ember-cli/ember-cli/blob/d209b88581bd2a723ab6e0fbac2f3b7f00a436b7/lib/broccoli/ember-app.js#L1616-L1649) `this` is not the `engine-addon`, but an `ember-app`. For addons it would be an [`ember-addon`](https://github.com/ember-cli/ember-cli/blob/d209b88581bd2a723ab6e0fbac2f3b7f00a436b7/lib/broccoli/ember-addon.js), which in turn extends the `ember-app`.

The way I see it, `this.env` is alsways defined. Indeed, if you try out my [fa-engine-demo](https://github.com/buschtoens/fa-engine-demo) and actually hook into the `ember-engines/lib/engine-addon.js` file, you can see that `target.env` is defined.

Is there any other problem with this PR, that I could fix? Because this is currently a show stopper for us. :sweat_smile:

rwjblue:
@buschtoens:

You are correct that `_getAssetPath` is a method on `ember-cli/lib/broccoli/ember-app`, however if you notice what is happening in the custom `import` method that we create [here](https://github.com/ember-engines/ember-engines/blob/97e6c831949ded0fc59c7bf0f875f25867171aa5/lib/engine-addon.js#L299-L304) you will see that we _copy_ the `_getAssetPath` method from `lib/broccoli/ember-app.js` instance, to the "host".  Our custom [findHost](https://github.com/ember-engines/ember-engines/blob/97e6c831949ded0fc59c7bf0f875f25867171aa5/lib/engine-addon.js#L56-L68) is ensuring that the `target` there is either the `lib/broccoli/ember-app.js` instance **OR** the nearest "lazy loading engine". Please note, that addons (or engines at this moment) do not have a `this.env` property set.

So, as you can see, my original comment in https://github.com/ember-engines/ember-engines/pull/252#issuecomment-272631037 is mostly correct (the only part that was wrong is that I was suggesting to make `this.env` a function, but its actually supposed to be a property with a string value).

rwjblue:
I just pushed https://github.com/ember-engines/ember-engines/pull/252/commits/6b9df1610dc9cf9e52b1f77df7a5bfa852ad59d2 to address my concern above.  

@trentmwillis - Ready for another round of review, when you have time...

lib/engine-addon.js
```diff
@@ -294,17 +294,22 @@ module.exports = {
...
         target._import = host._import;
         target._getAssetPath = host._getAssetPath;
         target.otherAssets = host.otherAssets;
         target._mergeTrees = host._mergeTrees;
 
         // We're delegating to the upstream EmberApp behavior for eager engines.
         if (this.lazyLoading !== true) {
+          // Asset can be an object with environment properties.
+          asset = target._getAssetPath(asset);
+
```
trentmwillis: I think `asset` can be `undefined` here depending on `env`, so we should guard for that and return early.

lib/engine-addon.js
```diff
@@ -294,17 +294,22 @@ module.exports = {
         return buildExternalTree.call(this);
       };
 
+
```
trentmwillis: Nit: Undo this addition

buschtoens:
@rwjblue That clears things up! Thanks for taking this on. :heart:

DegreeDev:
Any update on this?

devotox:
Will this be merged?

trentmwillis:
@ima007 any chance you'll be able to work on this again soon? If not, no worries, we can have someone else pick it up.

ima007:
Yup! Taking a look at it again within the next week. Thanks for the ping/reminder @trentmwillis

albertpak:
@ima007 - any update on this PR? Can we get this merged in soon?

buschtoens:
@ima007 can you get around to get this PR over the finish line? :pray: 

If not, would you be okay with me taking this over and making the necessary changes? :slightly_smiling_face:  
I'd really love to get this merged.

trentmwillis:
At this point, since it seems multiple folks are waiting for this, if someone else would like to pick this work up (maybe work of the commits here so both authors get credit?), I'd be glad to review a new PR.

buschtoens:
@trentmwillis I made the changes you requested in https://github.com/ember-engines/ember-engines/pull/252#pullrequestreview-25835676

Is this what you meant by "return early"?

ima007:
Thanks everyone (Sorry for dropping the ball).

Issue #257, opened 2016-11-14
[PERF] RollupFunnel is unstable


stefanpenner:
Although it is possible to implement naive broccoli-plugins, by merely rebuilding everything every time. This does not enable a performant build.

Typically an efficient plugin must:

* calculate a patchset: https://github.com/broccolijs/broccoli-funnel/blob/master/index.js#L304
* apply that patchset efficiently to a persistent output: https://github.com/broccolijs/broccoli-funnel/blob/master/index.js#L317-L319

---
In this case, it is likely sufficient to:

1. figure out what changed
2. apply changset to internal representation
3. apply additions/removals to satisfy the filter based on 2.

---

This does require some effort, some WIP existing [here](https://github.com/ember-cli/rfcs/pull/76) and [here](https://github.com/stefanpenner/fs-tree-diff/pull/36) to improve the ergonomics. Unfortunately, to implement on-trivial transforms efficiently requires extra care. On a plus side, it does enable power-users to get all the performance there hart desires.

It is unclear to my, why rollup is required here. As I believe the usage today is merely to extract deps, which can be done likely directly via some [ES parser](https://github.com/ef4/ember-browserify/blob/master/lib/stubs.js#L115-L135)

It is also not entirely clear, why folder structure + dependencies alone are not enough to infer the information. But there is maybe a featureset I am unaware of that supports this.

Anyways, A good place to get started would be to create a test ensure stable output, an example of such a test: https://github.com/eploko/broccoli-watchify/blob/master/test/watchify-test.js#L36-L62

trentmwillis:
This should be fixed now.

Issue #256, opened 2016-11-14
RollupFunnel missing tests


stefanpenner:
https://github.com/dgeb/ember-engines/blob/master/lib/rollup-funnel.js

Examples of how to test broccoli plugins

* https://github.com/chadhietala/broccoli-rollup/blob/master/tests/index.js
* https://github.com/eploko/broccoli-watchify/blob/master/test/watchify-test.js
* https://github.com/broccolijs/broccoli-funnel/blob/master/tests/index.js

trentmwillis:
I'm gonna work on this. Once the tests are in place we can work on #257 to ensure we avoid regressions.

stefanpenner:
@trentmwillis awesome. Yes having tests will leave a nice foundation for future contributions (or bug fixers)

trentmwillis:
Fixed and subsequently removed from this repo to a separate package.

PR #261, opened 2016-11-16
Add tests for RollupFunnel + make it stable

trentmwillis:
Addressing #256 & #257 

cc @stefanpenner @nathanhammond @rwjblue

### DIFF
--- a/lib/rollup-funnel.js
+++ b/lib/rollup-funnel.js
@@ -5,6 +5,9 @@ var Plugin = require('broccoli-plugin');
 var rollup = require('rollup').rollup;
 var amdNameResolver = require('amd-name-resolver').moduleResolve;
 var existsSync = require('exists-sync');
+var FSTree = require('fs-tree-diff');
+var Entry = require('fs-tree-diff/lib/entry');
+var rimraf = require('rimraf');
 
 function filter(directory, prefix, callback) {
   var results = [];
@@ -33,6 +36,22 @@ function copy(inodes) {
   }, this);
 }
 
+/**
+ * Stats a path. If it exists, it returns the stat information. Otherwise it
+ * returns null.
+ *
+ * @param {String} path
+ * @return {fs.Stats}
+ */
+function existsStat(path) {
+  try {
+    var stat = fs.statSync(path);
+    return stat;
+  } catch (e) {
+    return null;
+  }
+}
+
 // Create a subclass RollupFunnel derived from Plugin
 RollupFunnel.prototype = Object.create(Plugin.prototype);
 RollupFunnel.prototype.constructor = RollupFunnel;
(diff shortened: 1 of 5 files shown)


node-tests/rollup-funnel-test.js
```diff
@@ -0,0 +1,93 @@
...
+          external: [ 'ember-engines/routes' ]
+        }
+      });
+
+      var builder = new broccoli.Builder(rollupFunnel);
+
+      return builder.build().then(function(result) {
+        var output = walkSync(result.directory);
+        assert.deepEqual(output, [ 'engine.js', 'utils/', 'utils/herp.js' ]);
+        return builder.cleanup();
```
stefanpenner: should be called in `afterEach`

likely in the category of "an exploration of an abstraction to hide these details"

stefanpenner: https://github.com/chadhietala/broccoli-rollup/blob/master/tests/index.js#L18
https://github.com/chadhietala/broccoli-rollup/blob/master/tests/index.js#L40
https://github.com/chadhietala/broccoli-rollup/blob/master/tests/index.js#L35

trentmwillis: Generally, I dislike that pattern as it encourages state sharing, which can lead to test ordering problems. In this case, however, it is likely okay. Will just also suggest setting `builder` to `undefined` in `afterEach`.

trentmwillis:
@stefanpenner updated with your comments.

trentmwillis:
Updated this with an attempt at making rebuilds stable/performant. Not quite finished though.

These are the scenarios I am thinking of:

- No changes -> output is stable
- Using `include` and dep graph changes -> run rollup to recalculate output
- Using `include` and change occurs outside dep graph -> output is stable
- Using `exclude` and dep graph changes -> run rollup to recalculate output
- Using `exclude` and change occurs outside dep graph -> copy forward changes

lib/rollup-funnel.js
```diff
@@ -45,25 +47,95 @@ function RollupFunnel(inputNodes, options) {
...
+    var absolutePath = path.join(inputPath, entryPath);
+
+    if (!fs.existsSync(absolutePath)) {
+      return;
+    }
+
+    var stat = fs.lstatSync(absolutePath);
+    var isDirectory = stat.isDirectory();
+
+    return {
```
trentmwillis: I would appreciate this, or even having `fromPaths` do the stating. It was super unintuitive to me why `fromPaths` wasn't doing what I wanted it to do.

lib/rollup-funnel.js
```diff
@@ -45,25 +47,95 @@ function RollupFunnel(inputNodes, options) {
...
+        }, this);
+        return;
+      }
+    }
+  }
+
   var modules = [];
 
-  var entryExists = existsSync(path.join(this.inputPaths[0], this.options.rollup.entry));
+  var entryExists = existsSync(path.join(base, this.options.rollup.entry));
```
trentmwillis: Not intentional, just due to me overlooking it. I think I added all the `fs.existsSync`, so I'll update to use `exists-sync` instead.

lib/rollup-funnel.js
```diff
@@ -99,16 +171,26 @@ RollupFunnel.prototype.build = function() {
   return rollup(rollupOptions).then(function() {
     var toCopy;
 
+    this._depGraph = modules.sort();
+    this._nonDepGraph = filter(base, '', function(module) {
+      return !~modules.indexOf(module);
```
trentmwillis: Nah, I'll change it.

trentmwillis:
@rwjblue addressed most of your comments. I think this is in a pretty good place now.

rwjblue:
> Also, would we potentially want to extract this into a separate package? Would be nice to iterate on it independent of the ember-engines release cycle

Yes, I agree.

trentmwillis:
@rwjblue that was my plan

rwjblue:
This looks good to me, I'd like to land this and release so that we can continue to iterate (and possibly extract). 

Any objections?

PR #263, opened 2016-11-17
Document dependency setup for engines

knownasilya:
Talked about in slack. cc @nathanhammond

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,26 @@ your project, as described above.
 
 ### Configuring your Engine
 
+Engines depend on `ember-engines` as a peer dependency, which means
+that it must be installed in the consuming app. So make sure that it's
+specified as suck in your `package.json`:
+
+```json
+  "peerDependencies": {                                                                                                   
+    "ember-engines": "*" 
+  },
+```
+
+_Note: For engines that are stand alone addons, that is, not an in-repo-engine,
+make sure to also add `ember-engines` as a dev dependency. This allows local
+development._
+
+```json
+  "devDependencies": {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,26 @@ your project, as described above.
...
+that it must be installed in the consuming app. So make sure that it's
+specified as such in your `package.json`:
+
+```json
+  "peerDependencies": {                                                                                                   
+    "ember-engines": "*" 
+  },
+```
+
+_Note: For engines that are stand alone addons, that is, not an in-repo-engine,
```
rwjblue: Looks like a small typo here: s/aLone/alone/

knownasilya: I don't see it lol..

README.md
```diff
@@ -89,6 +89,25 @@ your project, as described above.
...
+that it must be installed in the consuming app. So make sure that it's
+specified as such in your engine's `package.json`:
+
+```json
+  "peerDependencies": {                                                                                                   
+    "ember-engines": "*" 
+  },
+```
+
+Also, make sure to add `ember-engines` as a dev dependency. This allows local
```
nathanhammond: This should specify into which `package.json` and it should. We should also clarify how to do this for in-repo vs. standalone engines.

knownasilya:
Updated

mydea:
I just created #307, and then saw this PR which does something like that. I'd like to have more information about the differences between devDependencies & dependencies, as that was not completely obvious to me (initially).

villander:
@dgeb @rwjblue this can be closed in favor this https://github.com/ember-engines/ember-engines.com/pull/46

PR #275, opened 2016-12-01
Replace rollup-funnel with broccoli-dependency-funnel

trentmwillis:
Should be good to go once we release [broccoli-dependency-funnel](https://github.com/ember-engines/broccoli-dependency-funnel)


trentmwillis:
I pushed before publishing broccoli-dependency-funnel, which is why AppVeyor is red (I don't have privs to kick it). Travis is passing thought and my local testing shows it is stable!

PR #276, opened 2016-12-03
Update Travis CI Badge

mike183:


trentmwillis:
Thank you! Would you mind submitting a similar PR for ember-asset-loader? I think it has the same issue.

mike183:
Will do, just noticed that there are also other references that need to be updated. Will submit another PR shortly.

PR #277, opened 2016-12-03
Update links in README

mike183:
There are still a couple of links under the 'Demo Projects' heading that reference repositories in @dgeb's account, not sure if these are being moved so I've left the links alone.


nathanhammond:
Thanks @mike183!

PR #281, opened 2016-12-13
Let ember-cli compile addon js

trentmwillis:
Instead of manually invoking `broccoli-babel-transpiler` we will allow either `compileAddon` or `originalTreeForAddon` (both provided by Ember-CLI) to do transpiling for us. This should work on existing versions of Ember-CLI and prevents us from breaking in upcoming versions.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -5,7 +5,7 @@ var existsSync = require('exists-sync');
 var fs = require('fs');
 var path = require('path');
 var writeFile = require('broccoli-file-creator');
-var babelTranspiler = require('broccoli-babel-transpiler');
+var Babel = require('broccoli-babel-transpiler');
 var concat = require('broccoli-concat');
 var DependencyFunnel = require('broccoli-dependency-funnel');
 var amdNameResolver = require('amd-name-resolver').moduleResolve;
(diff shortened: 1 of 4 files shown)


lib/engine-addon.js
```diff
@@ -66,7 +65,7 @@ function buildConfigTree(prefix) {
   // Include a module that reads the engine's configuration from its
   // meta tag and exports its contents.
   var configContents = this.getEngineConfigContents();
-  var configTree = writeFile(prefix + '/config/environment.js', configContents);
```
nathanhammond: Can you explain this one to me? Are we relying on the fact that it gets returned from the addon to namespace it?

trentmwillis: Yep, that's the gist of it. Gotta find a way to make these changes play nicely with non-master ember-cli though

lib/engine-addon.js
```diff
@@ -403,33 +400,21 @@ module.exports = {
           external: ['ember-engines/routes']
         });
 
-        // Babelify, but only to the extent of converting modules.
-        var babelOptions = {
```
rwjblue: Keeping this around as a fallback should make it compat with non-master ember-cli.

I'm thinking something like:

    var babelOptions = (this.options && this.options.babel) || { /* current pojo here */}
      
    }

lib/engine-addon.js
```diff
@@ -66,7 +76,7 @@ function buildConfigTree(prefix) {
   // Include a module that reads the engine's configuration from its
   // meta tag and exports its contents.
   var configContents = this.getEngineConfigContents();
-  var configTree = writeFile(prefix + '/config/environment.js', configContents);
+  var configTree = writeFile('config/environment.js', configContents);
```
nathanhammond: Can drop the `prefix` arg.

PR #283, opened 2016-12-19
Leverage addon tree caching from Ember-CLI canary

trentmwillis:


lib/engine-addon.js
```diff
@@ -10,7 +10,9 @@ var concat = require('broccoli-concat');
 var DependencyFunnel = require('broccoli-dependency-funnel');
 var amdNameResolver = require('amd-name-resolver').moduleResolve;
 var defaultsDeep = require('lodash.defaultsdeep');
+var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 
+var ADDON_TREE_CACHING = require('ember-cli/lib/experiments').ADDON_TREE_CACHING;
 var EOL = require('os').EOL;
```
rwjblue: Needs to try/catch since `lib/experiments` isn't present on non-master ember-cli versions

rwjblue: I'd suggest doing this:

```js
var ADDON_TREE_CACHING = (function() {
try {
    var experiments = require('ember-cli/lib/experiments');

    return experiments.ADDON_TREE_CACHING || 'cacheKeyForTree';
  } catch(e) {
    // just return `cacheKeyForTree` regardless 
    // but it won't be called
    return 'cacheKeyForTree';
  }
})();
```

lib/engine-addon.js
```diff
@@ -10,7 +10,16 @@ var concat = require('broccoli-concat');
...
 var defaultsDeep = require('lodash.defaultsdeep');
-
+var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
+var Addon = require('ember-cli/lib/models/addon');
+
+var ADDON_TREE_CACHING = (function() {
+  try {
+    return require('ember-cli/lib/experiments').ADDON_TREE_CACHING;
+  } catch (e) {
+    return null;
```
rwjblue: Sorry for the run around here, but I think we should tweak this slightly:

```js
try {
  return require('ember-cli/lib/experiments').ADDON_TREE_CACHING;
} catch(e) {
  if (e.code === 'MODULE_NOT_FOUND') {
    return null;
  }

  throw e;
}
```

But I'm happy to tweak this after landing, totally up to you...

PR #288, opened 2016-12-20
Fix current-when behaviour within an engine

### LINKED ISSUE #285: Inconsistency with link-to component
The route passed into the link-to component does not require the engine name (relative) but in order for current-when to work properly, you need to pass in the engine name + route (absolute). This seems like an inconsistency that should be changed. 

Example:
```{{#link-to 'route' current-when='engine.route'}}```

Ember 2.10.1
Engines 0.4.0

### PR DESCRIPTION
**Pull request by mike183**
This PR resolves the issue described in #285 and also includes a test that was previously failing. While the test does work, I do feel like there may be a better way to test this, possibly in an integration test?

Fixes #285

### DIFF
--- a/addon/-private/link-to-component.js
+++ b/addon/-private/link-to-component.js
@@ -15,9 +15,16 @@ export default LinkComponent.extend({
     let owner = getOwner(this);
 
     if (owner.mountPoint) {
+      // Prepend engine mount point to targetRouteName
       let fullRouteName = owner.mountPoint + '.' + get(this, 'targetRouteName');
-
       set(this, 'targetRouteName', fullRouteName);
+
+      // Prepend engine mount point to current-when if set
+      let currentWhen = get(this, 'current-when');
+      if (currentWhen !== null) {
+        let fullCurrentWhen = owner.mountPoint + '.' + get(this, 'current-when');
+        set(this, 'current-when', fullCurrentWhen);
+      }
     }
   }
 });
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: addon/-private/link-to-component.js [hunk 1]
```diff
@@ -15,9 +15,16 @@ export default LinkComponent.extend({
...
     if (owner.mountPoint) {
+      // Prepend engine mount point to targetRouteName
       let fullRouteName = owner.mountPoint + '.' + get(this, 'targetRouteName');
-
       set(this, 'targetRouteName', fullRouteName);
+
+      // Prepend engine mount point to current-when if set
+      let currentWhen = get(this, 'current-when');
+      if (currentWhen !== null) {
+        let fullCurrentWhen = owner.mountPoint + '.' + get(this, 'current-when');
```
trentmwillis: Should just use `currentWhen` instead of doing `get()` again.

tests/dummy/lib/eager-blog/addon/templates/post.hbs
```diff
@@ -0,0 +1,3 @@
+<div>
```
trentmwillis: Can we simplify this template to just be an `h4` or just `p`? Minor, but want the markup to be semantic.

addon/-private/link-to-component.js
mike183: Not sure how I missed this...I explicitly created that variable so that `get()` didn't have to be executed multiple times. Will update now.

mike183:
Thanks for the review, made the requested changes.

mike183:
Tests seem to be failing, not sure why? Is it related to the testing changes that were made yesterday?

trentmwillis:
Yeah, the only tests failing are due to the build diff changing (need to improve that). The actual Ember tests are passing.

Merging this. Thanks!

PR #289, opened 2016-12-20
Build styles tree before building public tree.

whastings:
This change allows non-CSS assets to be gathered during the building of the styles tree and then later added to the public tree. For example, [ember-cli-eyeglass](https://github.com/sass-eyeglass/ember-cli-eyeglass) needs this so it can add any non-CSS assets returned from Eyeglass to the public tree.

This also makes the process for building a lazy engine consistent with how Ember CLI builds regular addons, in that it builds the styles tree before the public tree.


whastings:
Also, I added a dummy engine to engine-testing for verifying this change: https://github.com/ember-engines/engine-testing/pull/3

Issue #287, opened 2016-12-20
Testing the EngineAddon build system


trentmwillis:
It is the unfortunate reality that the [`EngineAddon` class](https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js) is currently untested. This makes it difficult to improve/refactor the build system due to lack of confidence in changes.

In order to test the build system we should automate usage of the [`engine-testing` application](https://github.com/ember-engines/engine-testing) that we have, up till now, been using to manually build and verify the correctness of the build. This build should incorporate the following scenarios:

- eager engine
- eager engine > eager engine
- eager engine > lazy engine
- lazy engine
- lazy engine > eager engine
- lazy engine > lazy engine

Each of the above scenarios should also include additional addon dependencies to verify they are correctly included into the built output.

Using that build, the test we run should do the following:

1. Build engine-testing against the current `master` branch of `ember-engines`
2. Build engine-testing with changes applied to `ember-engines`
3. Compare both build outputs

This testing approach assumes that the current implementation is producing "correct" output. It will give us a high-level acceptance test initially. After landing the initial test, we should also support modifying the dependency graph of the `engine-testing` app such that we can build with only one of the above scenarios so that we can get more granularity in regards to possible issues.

We should likely avoid unit testing since the algorithmic approach of the `EngineAddon` class is likely subject to change since it was never thoroughly designed. The usage of acceptance tests will allow us to have confidence in compatibility with current consumers while not hindering refactors of the build process.

cc @nathanhammond @rwjblue

trentmwillis:
We have basic testing in place now (via #290); allowing us to iterate on the existing behavior.

trentmwillis:
Reopening. We should set up a custom Travis build script to only run the build tests when code in `index.js` or `lib/` has changed. Otherwise, we get failures even when they aren't "legit".

trentmwillis:
Fixed via https://github.com/ember-engines/ember-engines/pull/296.

PR #290, opened 2016-12-20
Add build acceptance test

trentmwillis:
Initial pass at addressing: https://github.com/ember-engines/ember-engines/issues/287

It's worth noting that with this approach, if we intentionally change the build output, then we will have a failing test. I think this is fine as it will force us to acknowledge when we're changing the semantics of the build.

### DIFF
--- a/node-tests/acceptance/build-test.js
+++ b/node-tests/acceptance/build-test.js
@@ -1,0 +1,31 @@
+var path = require('path');
+var expect = require('chai').expect;
+var compareBuild = require('../helpers/compare-build');
+
+describe('Acceptance', function() {
+  describe('build', function() {
+    this.timeout(300000);
+
+    var build;
+
+    beforeEach(function() {
+      build = compareBuild();
+    });
+
+    afterEach(function() {
+      build.cleanup();
+    });
+
+    it('correctly builds engine-testing application', function() {
+      try {
+        // Verify the diff of before/after is empty
+        expect(build.diff.length).to.equal(0);
+      } catch (e) {
+        // If there is a diff, log it out to help debugging
+        console.log('BUILD DIFF');
+        console.log(build.diff);
+        throw e;
+      }
+    });
+  });
+});
--- a/node-tests/helpers/compare-build.js
+++ b/node-tests/helpers/compare-build.js
@@ -1,0 +1,49 @@
+var path = require('path');
+var execSync = require('execa').shellSync;
+
+function execute(command) {
+  console.info('Executing: ' + command);
+  return execSync(command);
+}
+
+/**
+ * Builds the engine-testing application against the current upstream version
+ * of ember-engines and against the current local copy. The return value is the
+ * list of files produced by both outputs so that we can catch build differences
+ * in testing.
+ */
+module.exports = function compareBuild() {
+  // clone engine-testing without the full git history
+  execute('git clone --depth 1 https://github.com/ember-engines/engine-testing.git');
+
+  // link ember-engines local
+  execute('npm link');
+
+  // cd into engine-testing
+  var originalCWD = process.cwd();
+  var engineTestingCWD = path.join(originalCWD, 'engine-testing');
+
+  process.chdir(engineTestingCWD);
+
+  // install all dependencies
+  execute('npm install');
+  execute('bower install');
+
+  // build against current master of ember-engines (upstream)
+  execute('ember build --output-path upstream-build');
+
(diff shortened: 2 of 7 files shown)


node-tests/helpers/compare-build.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+module.exports = function compareBuild() {
+  // clone engine-testing
+  execute('git clone https://github.com/ember-engines/engine-testing.git');
+
+  // link ember-engines local
+  execute('npm link');
+
+  // cd into engine-testing
+  process.chdir('engine-testing');
```
nathanhammond: Let's specify this via `path.join(original, engine-testing)`

node-tests/helpers/compare-build.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Builds the engine-testing application against the current upstream version
+ * of ember-engines and against the current local copy. The return value is the
+ * list of files produced by both outputs so that we can catch build differences
+ * in testing.
+ */
+module.exports = function compareBuild() {
+  // clone engine-testing
+  execute('git clone https://github.com/ember-engines/engine-testing.git');
```
nathanhammond: Don't need a clone:

```
wget https://github.com/ember-engines/engine-testing/archive/master.zip
unzip engine-testing-master.zip
mv engine-testing-master engine-testing
```

Then again, cross-platform means that has to happen in `node`. Gross. Nevermind, just do a shallow clone.

trentmwillis:
Addressed @nathanhammond's comments and opened #291 to show that this test catches build changes.

node-tests/acceptance/build-test.js
```diff
@@ -0,0 +1,41 @@
...
+
+      // Verify the before and after builds have the same files
+      expect(newBuildFiles).to.deep.equal(oldBuildFiles);
+
+      // Expect the contents of all the build files to be the same
+      for (var i = 0; i < oldBuildFiles.length; i++) {
+        var oldFilePath = path.join(build.oldFilesPath, oldBuildFiles[i]);
+        var newFilePath = path.join(build.newFilesPath, newBuildFiles[i]);
+
+        expect(file(newFilePath)).to.equal(file(oldFilePath));
```
nathanhammond: Were you going to use `git diff` and build artifacts for this?

nathanhammond: Were you going to use `git diff` and build artifacts for this?

trentmwillis: `git diff`, yes; build artifacts, no. You have to upload artifacts somewhere external to Travis and I don't feel that hassle is worth it.

trentmwillis:
@nathanhammond updated this and #291 to use `git diff` instead. I'm comfortable with landing this now. I'd like to improve some ergonomics of working with it locally, but that can be done in a follow up.

package.json
```diff
@@ -47,8 +51,11 @@
     "ember-load-initializers": "^0.5.1",
     "ember-resolver": "^2.1.0",
     "ember-sinon": "0.5.0",
+    "execa": "^0.5.0",
     "fs-extra": "^1.0.0",
-    "loader.js": "^4.0.1"
+    "loader.js": "^4.0.1",
+    "mocha": "^3.2.0",
+    "walk-sync": "^0.3.1"
```
nathanhammond: Can be removed.

package.json
```diff
@@ -27,6 +29,8 @@
   "license": "MIT",
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.2",
+    "chai": "^3.5.0",
+    "chai-files": "^1.4.0",
```
nathanhammond: Can be removed.

nathanhammond:
We have a problem: https://travis-ci.org/ember-engines/ember-engines/jobs/185693824#L389

Because the entire file is one line the diff is useless and Travis is blowing up on us. :/

trentmwillis:
> Because the entire file is one line the diff is useless and Travis is blowing up on us. :/

Yes, but the only reason that file is different is because the files preceding it in the logs had diffs as well. I think the diff up till that point provides enough feedback to be actionable.

trentmwillis:
> we should definitely continue refactoring so that we can get more meaningful and fine-grained tests in place...

Most definitely. In fact, the primary goal of this test is so that we can refactor more quickly and with better confidence so that we can get to a place that we have more "standard" tests.

PR #292, opened 2016-12-21
Align lazy and eager builds

trentmwillis:
In order to more effectively cache builds and have an easier to understand build process, we want to align lazy and eager (non-lazy) build processes. Right now the two types of builds do a lot of similar steps, but don't share any underlying implementation, this PR will seek to share as much implementation as is reasonably possible.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -156,6 +156,15 @@ function buildEngineJSTree() {
   return engineTreeRelocated;
 }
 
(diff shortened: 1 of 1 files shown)


trentmwillis:
Gonna go ahead and merge this. If Nathan has any objections we can fix forward/revert if needed.

nathanhammond:
Probably broke this: https://github.com/ember-engines/engine-testing/pull/9 (linking these together)

PR #296, opened 2016-12-22
Run build tests only if appropriate changes were made

trentmwillis:
Updates Travis to only run build tests when `lib/*` or `index.js` are changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ cache:
 env:
   - TEST_SCRIPT=test:ember EMBER_TRY_SCENARIO=ember-canary
   - TEST_SCRIPT=test:ember EMBER_TRY_SCENARIO=ember-beta
-  - TEST_SCRIPT=test:node
+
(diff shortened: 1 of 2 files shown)


rwjblue:
This is great work, how would you feel about inverting the guard?  I'm specifically thinking to ignore changes to `addon/` and `app/` but run  `test:node` for everything else...

trentmwillis:
> how would you feel about inverting the guard? I'm specifically thinking to ignore changes to `addon/` and `app/`

This seems like a much better solution to me. Will update with this approach.

trentmwillis:
@rwjblue updated this, but didn't invert the guard. Thinking about this further, we would need to check `app/`, `addon/`, `blueprints/` and `tests/` (and soon `addon-test-support`). There are also other files which shouldn't impact downstream builds, so I think whitelisting is better. I think `lib/`, `node-tests/`, `index.js` and `package.json` should be good enough.

PR #300, opened 2016-12-26
Adds information about customising assets URI

hyderali:
Thought that this will be helpful. Took a considerable amount of time for me to figure out this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,8 +456,51 @@ behavior.
 
 Further, the engine must enumerate its primary assets (JS and CSS) in order to
 be loaded by the asset loading service. That will be generated at
-`/dist/asset-manifest.json` at build time. It will also by default be inserted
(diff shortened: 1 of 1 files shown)


trentmwillis:
Thanks for the PR @hyderali! Most of this information overlaps with a recent update to the ember-engines website guide: http://ember-engines.com/guide/deploying-engines. Can you see if those guides are missing anything you're adding here?

I would actually like to remove a large portion of the content in this `README` in favor of pointing folks to ember-engines.com so that there is a single source of truth. @dgeb / @rwjblue thoughts?

rwjblue:
Yes, the readme should be a small funnel to point folks at the website (for users) or contributing.md (for contribs)...

hyderali:
@trentmwillis , yeah all information required are there in the guides itself. Will close this PR. Thanks for looking into

PR #302, opened 2016-12-26
Memoize build functions

trentmwillis:
We have a lot of small functions that are potentially called more than once for an Engine. By caching the return values of these functions, we can improve build performance by not doing redundant work.

The memoization done here assumes that only one output value ever occurs for the function, which is an assumption currently made by the build system.


rwjblue:
👍 - I like the reuse of calculate-cache-key-for-tree 😎

nathanhammond:
This PR is the cause of the bug here: https://github.com/ember-engines/engine-testing/pull/5

I'm in the process of identifying which of the newly `memoize`d functions is triggering this error.

nathanhammond:
Note: this is _not_ the commit which introducing the aforementioned problem; but it is likely introducing problems nonetheless for the functions which receive arguments.

Issue #274, opened 2016-11-28
Confirm that all engine assets appear in `preprocessTree` and `postprocessTree`.


nathanhammond:
I _believe_ that lazy engine CSS/JS/template assets are not being pushed into `preprocessTree` and `postprocessTree` as a result of being handled inside of `treeForPublic`. This means that addons which register hooks for those don't receive the contents they're expecting.

Likely need to manually invoke those hooks with engine assets, though it does mean that the hook may run multiple times. This could possibly be considered breaking.

rwjblue:
> All assets are present (lazy, eager, and imported) in assets/vendor.css.

I believe we got somewhat confused on this one, the only assets present in `assets/vendor.css` were those using `app.import` (`this.import` was handled correctly as was thigns in `addon/styles` of the engine or `addon/styles` of nested child addons).  https://github.com/ember-engines/ember-engines/pull/314 fixes the `app.import` issue.

> [BUG] CSS preprocessTree is being invoked twice.

PR has been submitted to address this https://github.com/ember-cli/ember-cli/pull/6676.

> [BUG] pre/postprocessTree is only invoked one-level-deep.

Has been fixed upstream in https://github.com/ember-cli/ember-cli/pull/6516 and will be released in ember-cli@2.12 (first beta soon).

> [BUG] The app folder isn't fully nuked inside of engines.

I'd like to separate that into a separate issue, we should likely prevent engines from having `app/` folders.

> this.import('vendor/addon.css') has all CSS in the entire hierarchy, we thought we noop'd it. Why?

Same as the first issue, `app.import` was misbehaving, and should be fixed by https://github.com/ember-engines/ember-engines/pull/314.

PR #314, opened 2017-01-14
Prevent hoisting of all `app.import`'s.

rwjblue:
Before this change, every addon that did an `app.import` on the argument passed in to the `included` hook would end up hoisting the thing imported all the way to the top.

This change passes in the engine itself as the `app` so that it is properly handled (addons downstream from a lazy engine end up in the lazy engines `engine-vendor.{js,css}` and addons downstream from eager engines end up in the top level `vendor.{js,css}`).

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -311,7 +311,7 @@ module.exports = {
           `this.parent`.
          */
         if (originalIncluded) {
-          originalIncluded.call(this, originalFindHost.call(this));
+          originalIncluded.call(this, this);
         }
 
         this.eachAddonInvoke('included', [this]);


rwjblue:
Tests are failing due to the various output files changing contents, I will create a better diff so that we can review the actual changes to confirm this is functioning properly.

trentmwillis:
> This change passes in the engine itself as the app so that it is properly handled

Just to make sure I understand, this works the way you described because we [define `this.import`](https://github.com/rwjblue/ember-engines/blob/554c6eeab7615b1d7c0283d77a834c073b50d78c/lib/engine-addon.js#L281-L295) for Engines, correct?

Overall, this seems fine. A cleaned up diff would be nice and a running of `ember test` in `engine-testing`, but after that I'm :+1: to merge.

rwjblue:
> this works the way you described because we define this.import for Engines, correct?

confirm, exactly

> A cleaned up diff would be nice

Yes, the diff in the output on Travis is not helpful and we still need to sanity check.

rwjblue:
OK, better diff is here: https://github.com/rwjblue/engine-testing-dist-output/pull/1

You can see that the only changes are moving things that were supposed to be lazy from `assets/vendor.js` into the respective `engines-dist/` location.

Also, confirmed `ember test` passes in engine-testing after this (though there really aren't any tests there other than ESLint).

rwjblue:
I think we will have to merge over the failing Travis build (since its failing due to assets changing, which is intentional) so I don't want to do that without @trentmwillis and/or @nathanhammond having a chance to review the diff in https://github.com/rwjblue/engine-testing-dist-output/pull/1.

trentmwillis:
Looks good to me.

> confirmed ember test passes in engine-testing after this (though there really aren't any tests there other than ESLint).

That's fine, it's mainly just to make sure everything still loads up properly.

> I think we will have to merge over the failing Travis build

Yep. This is fine, but it does force us to acknowledge we are changing the build semantics a bit. Though in this case it should be considered a bug.

nathanhammond:
Hold on a second!

trentmwillis:
Oops, sorry @nathanhammond. If you see something wrong, let me know and we can revert.

nathanhammond:
I did this _very_ intentionally to maximize compatibility of `ember-engines` with the existing addon community. I noted a ton of places where `app.import` behavior inside of a nested addon relied on the top-level hoisting and would result in really strange or broken behavior–as in, the assets would not be present at the right time or place. (This mattered more for JS resources than it did CSS resources.) I spent a lot of time refactoring the engines build to support both `.import` behaviors and tested against numerous addons.

So, there is already branching in the codebase to support doing the "right" thing for `this.import`–as evidenced by this single line change (there are other ways to change this that involve removing the branching). However, rather than breaking compat with people blissfully ignorant and using `app.import` I'd rather we keep this behavior. But! We need to move people to the "right" solution: I'd like to propose that we warn on the console instead.

I'm going to press the "Revert" button as I already know that this change is less community compatible. I think that simply providing the warning will allow early adopters to start moving their addons and the rest of the community to the ideal solution (`this.import`).

rwjblue:
I'm happy to add a warning here instead.  The main thing that this shows is exactly what is causing the behaviors that both you and I perceived as fundamentally broken when we opened #274.  Now that we understand the cause (see my last comment there) I think warning is totally fine.

nathanhammond:
I'm so sad that I forgot about this feature I added. 😞

PR #324, opened 2017-01-17
Ensure `preprocessTree` / `postprocessTree` are invoked properly for lazy engine styles.

### LINKED ISSUE #322: Brainstorm session on unifying app and engine style pipelines...
After a pairing session with @stefanpenner which mostly focused on aligning the styles generation, hooks being called, etc we came up with the following proposed result.  Still very much WIP and somewhat still in my brain, but I'll be working on this over the next day or so...

---

Normal App (also runs for eager engines) 
----------
1. build up addonTreesFor('styles')
1. process `app/styles`
1. merge trees from #1 and #2 above.
1. invoke preprocessTree('css', ...)
1. process registry of CSS preprocessors
1. concat into `options.outputPaths.app.css`
1. process things that were `app.import`'ed (e.g. `assets/vendor.css`)
1. prod ? preprocessMinifyCss : null
1. invoke postprocessTree('css', ...)

Lazy Engines (today)
------------
1. `this.compileStyles(this._treeFor('addon-styles'));`
  * preprocessTree('css', ...)
  * `processCss` (main registry processors e.g. sass)
  * postprocessTree('css', ...)
1. concat `**/*.css` -> `engines-dist/<engine-name>/assets/engine.css`
1. `this.eachAddonInvoke('treeFor', ['addon'])` as `vendor`
1. concat `**/*.css` from vendor -> `engine-dist/<engine-name>/assets/engine-vendor.css`
1. process `app.import`ed styles

Lazy Engines (Proposed)
----------
1. build up addonTreesFor('styles') *Note: not required for today*
1. process `treeFor('addon-styles');`
1. merge trees from #1 and #2 above
1. invoke preprocessTree('css', ...)
1. concat `**/*.css` -> `engines-dist/<engine-name>/assets/engine.css`
1. `this.eachAddonInvoke('treeFor', ['addon'])` as `vendor`
1. process things that were `app.import`'ed (e.g. `assets/vendor.css`)
1. prod ? preprocessMinifyCss : null *Note: unclear if this is needed at the lazy engine level AND the app level?*
1. invoke postprocessTree('css', ...)

rwjblue:
New / tweaked styles pipeline is:

1. Invoke `preprocessTree` on all of the engines dependencies, with the raw `addon/styles` tree.
2. Process the raw tree through all the preprocessors setup in the registry.
3. Concat `**/*.css` from the result of 2 into `engine.css`
4. Concat vendor styles (dependent addons' `addon-styles` trees) into `engine-vendor.css`
5. Process `app.import`'s related to styles.
6. Invoke `postprocessTree` on all of the engines dependencies, with the final combined output of 3, 4, and 5.

Fixes #322

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -228,6 +228,10 @@ var buildCompleteJSTree = memoize(function buildCompleteJSTree() {
   );
 });
 
+var buildEngineStyleTree = memoize(function buildEngineStyleTree() {
+  return this._treeFor('addon-styles');
+});
+
 module.exports = {
   extend: function(options) {
     var originalInit = options.init || function() { this._super.init.apply(this, arguments); };
@@ -362,7 +366,7 @@ module.exports = {
 
         // NOT LAZY LOADING!
         // This is the scenario where we want to act like an addon.
-        var engineCSSTree = this._treeFor('addon-styles');
+        var engineCSSTree = buildEngineStyleTree.call(this);
         var compiledEngineCSSTree = this.compileStyles(engineCSSTree);
 
         // If any of this engines ancestors are lazy we need to
@@ -396,7 +400,7 @@ module.exports = {
         var engineStylesOutputDir = 'engines-dist/' + this.name + '/assets/';
 
         // Get base styles tree.
-        var engineStylesTree = this._treeFor('addon-styles');
+        var engineStylesTree = buildEngineStyleTree.call(this);
 
         var primaryStyleTree = null;
         if (engineStylesTree) {
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1,6 +1,6 @@
 var Funnel = require('broccoli-funnel');
 var merge = require('lodash/merge');
-var mergeTrees = require('broccoli-merge-trees');
+var mergeTrees = require('ember-cli/lib/broccoli/merge-trees');
 var existsSync = require('exists-sync');
 var fs = require('fs');
 var path = require('path');
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -13,6 +13,9 @@ var defaultsDeep = require('lodash.defaultsdeep');
 var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 var Addon = require('ember-cli/lib/models/addon');
 var memoize = require('./utils/memoize');
+var p = require('ember-cli-preprocess-registry/preprocessors');
+var preprocessJs = p.preprocessJs;
+var preprocessCss = p.preprocessCss;
 
 // required until stable version of ember-cli includes
 // https://github.com/ember-cli/rfcs/pull/90
@@ -166,10 +169,6 @@ var buildEngineJSTreeWithoutRoutes = memoize(function buildEngineJSTreeWithoutRo
     entry: this.name + '/routes.js',
     external: ['ember-engines/routes']
   });
-});
-
-var buildEngineCSSTree = memoize(function buildEngineCSSTree() {
-  return this.compileStyles(this._treeFor('addon-styles'));
 });
 
 var buildVendorJSWithImports = memoize(function buildVendorJSWithImports(concatTranspiledVendorJSTree) {
@@ -358,14 +357,15 @@ module.exports = {
 
       // Replace `treeForAddon` so that we control how this engine gets built.
       // We may or may not want it to be combined like a default addon.
-      this.treeForAddon = function(engineSourceTree) {
+      this.treeForAddon = function() {
         if (this.lazyLoading === true) { return; }
 
         // NOT LAZY LOADING!
         // This is the scenario where we want to act like an addon.
-        var engineCSSTree = buildEngineCSSTree.call(this);
+        var engineCSSTree = this._treeFor('addon-styles');
+        var compiledEngineCSSTree = this.compileStyles(engineCSSTree);
 
-        // If this engine is lazy or any of its parents are lazy we need to
+        // If any of this engines ancestors are lazy we need to
         // remove its routes, because we've already accounted for our routes
         // with the `treeForEngine` hook.
         var engineJSTree = this._hasLazyAncestor ?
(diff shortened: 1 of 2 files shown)


rwjblue:
The "diff tests" are failing due to a mismatched `ember-cli` version when ember-engines is linked within engine-testing.  There is actually no change in output before/after this PR.  The output is exactly the same as [this commit in engine-testing-dist-output](https://github.com/rwjblue/engine-testing-dist-output/tree/bdf714583010dee93e7c00248cee02cf92afc4a6).

lib/engine-addon.js
```diff
@@ -387,10 +387,76 @@ module.exports = {
...
+          outputFile: engineStylesOutputDir + 'engine-vendor.css'
+        });
+
+        var concatMergedVendorCSSTree = mergeTrees([concatVendorCSSTree, externalTree]);
+
+        // So, this is weird, but imports are processed in order.
+        // This gives the chance for somebody to prepend onto the vendor files.
+        var vendorCSSImportTree = buildVendorCSSWithImports.call(this, concatMergedVendorCSSTree);
+
+        var mergedVendorCSSWithImportAndEngineStylesTree = mergeTrees(
```
stefanpenner: do we need https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/merge-trees.js ?

lib/engine-addon.js
```diff
@@ -387,10 +387,76 @@ module.exports = {
         return mergeTrees([
           externalTree,
           engineJSTree,
-          engineCSSTree
+          compiledEngineCSSTree
         ].filter(Boolean), { overwrite: true });
       };
 
+      this.compileLazyEngineStyles = function compileLazyEngineStyles(vendorTree, externalTree) {
+        var vendorCSSTree = buildVendorCSSTree.call(this, vendorTree);
```
rwjblue: @stefanpenner - Yes, they are `memoize`ed with a custom `memoize` util function already.  We can clean it up a bit and make sure they are cached in the same way in future refactors.

lib/engine-addon.js
```diff
@@ -387,10 +387,76 @@ module.exports = {
...
+        var engineStylesOutputDir = 'engines-dist/' + this.name + '/assets/';
+
+        // Get base styles tree.
+        var engineStylesTree = this._treeFor('addon-styles');
+
+        var primaryStyleTree = null;
+        if (engineStylesTree) {
```
rwjblue: > I suspect it may have to do with actually needing to keep vendor + add-on style separate?

Confirm, that is exactly right. If we merge them, then do the `concat` for `**/*.css` we loose the ability to actually have separate `engine.css` and `engine-vendor.css`.

lib/engine-addon.js
rwjblue: Yes, I think we do (to help eliminate wasted merges and whatnot).  I'll update.

rwjblue:
Latest commit bumps `ember-cli` version to match what engine-testing uses, which fixes the node:ci tests (and confirms the built asset output is identical).

lib/engine-addon.js
```diff
@@ -387,10 +387,76 @@ module.exports = {
...
+          compiledEngineCSSTree
         ].filter(Boolean), { overwrite: true });
       };
 
+      this.compileLazyEngineStyles = function compileLazyEngineStyles(vendorTree, externalTree) {
+        var vendorCSSTree = buildVendorCSSTree.call(this, vendorTree);
+        var engineStylesOutputDir = 'engines-dist/' + this.name + '/assets/';
+
+        // Get base styles tree.
+        var engineStylesTree = this._treeFor('addon-styles');
```
trentmwillis: Should we memoize/cache the result of `_treeFor('addon-styles')`? Since it is called in more than one location.

rwjblue: Its only called once per engine instance, the first invocation is for non-lazy engines and the second is for lazy engines.  I think we can definitely swap to memoize though, will update.

rwjblue:
@trentmwillis - Should be ready for another look...

lib/engine-addon.js
```diff
@@ -358,14 +361,15 @@ module.exports = {
...
         if (this.lazyLoading === true) { return; }
 
         // NOT LAZY LOADING!
         // This is the scenario where we want to act like an addon.
-        var engineCSSTree = buildEngineCSSTree.call(this);
+        var engineCSSTree = buildEngineStyleTree.call(this);
+        var compiledEngineCSSTree = this.compileStyles(engineCSSTree);
 
-        // If this engine is lazy or any of its parents are lazy we need to
+        // If any of this engines ancestors are lazy we need to
```
alexander-alvarez: `engine's`

lib/engine-addon.js
```diff
@@ -358,14 +361,15 @@ module.exports = {
...
         // remove its routes, because we've already accounted for our routes
```
alexander-alvarez: `its` being the engine or the parent's ?

lib/engine-addon.js
```diff
@@ -387,10 +391,76 @@ module.exports = {
...
+        var vendorCSSTree = buildVendorCSSTree.call(this, vendorTree);
+        var engineStylesOutputDir = 'engines-dist/' + this.name + '/assets/';
+
+        // Get base styles tree.
+        var engineStylesTree = buildEngineStyleTree.call(this);
+
+        var primaryStyleTree = null;
+        if (engineStylesTree) {
+          var preprocessedEngineStylesTree;
+          // `_addonPreprocessTree` was added in Ember 2.12 to ensure
```
nathanhammond: Ember CLI 2.12.

lib/engine-addon.js
```diff
@@ -358,14 +361,15 @@ module.exports = {
...
       // Replace `treeForAddon` so that we control how this engine gets built.
       // We may or may not want it to be combined like a default addon.
-      this.treeForAddon = function(engineSourceTree) {
+      this.treeForAddon = function() {
         if (this.lazyLoading === true) { return; }
 
         // NOT LAZY LOADING!
         // This is the scenario where we want to act like an addon.
-        var engineCSSTree = buildEngineCSSTree.call(this);
+        var engineCSSTree = buildEngineStyleTree.call(this);
```
nathanhammond: This is now returning the nested already-preprocessed child CSS trees via https://github.com/ember-cli/ember-cli/blob/16e4492c9ebf3348eb0f31df17215810674dbdf6/lib/models/addon.js#L781-L794

You split it to enable memoization. Can some of this end up documented in comments?

lib/engine-addon.js
```diff
@@ -387,10 +391,76 @@ module.exports = {
...
           externalTree,
           engineJSTree,
-          engineCSSTree
+          compiledEngineCSSTree
         ].filter(Boolean), { overwrite: true });
       };
 
+      this.compileLazyEngineStyles = function compileLazyEngineStyles(vendorTree, externalTree) {
+        var vendorCSSTree = buildVendorCSSTree.call(this, vendorTree);
+        var engineStylesOutputDir = 'engines-dist/' + this.name + '/assets/';
```
nathanhammond: I'd prefer `path.join` unless `preprocessCss` expects POSIX in which case `ensure-posix-path`. We currently have to know whether we've included leading/trailing slashes anywhere we modify it such as https://github.com/ember-engines/ember-engines/pull/324/files#diff-1c2c32318ef6ba8d9be4e7b085f8ecb5R419

lib/engine-addon.js
```diff
@@ -406,21 +476,15 @@ module.exports = {
...
-        // All we need to do is concatenate it down.
-        var primaryStyleTree;
-        if (engineStylesTree) {
-          primaryStyleTree = concat(engineStylesTree, {
-            allowNone: true,
-            inputFiles: ['**/*.css'],
-            outputFile: 'engines-dist/' + this.name + '/assets/engine.css'
-          });
-        }
+        var finalStylesTree = this.compileLazyEngineStyles(vendorTree, externalTree);
```
nathanhammond: I'm concerned that this doesn't prevent hoisting of nested lazy engine's styles but haven't traced through it. I believe that this case isn't covered in our fixtures.

Given: `root > lazy1 > lazy2 > eager`

We would expect:
- `root.css` to have only its own assets.
- `lazy1/engine-vendor.css` to be empty.
- `lazy2/engine-vendor.css` to contain `eager`'s styles.

lib/engine-addon.js
```diff
@@ -387,10 +391,76 @@ module.exports = {
...
+
+        // Get base styles tree.
+        var engineStylesTree = buildEngineStyleTree.call(this);
+
+        var primaryStyleTree = null;
+        if (engineStylesTree) {
+          var preprocessedEngineStylesTree;
+          // `_addonPreprocessTree` was added in Ember 2.12 to ensure
+          // addons had preprocessTree / postprocessTree invoked properly
+          // this emulates the normal behavior before and after that version
```
nathanhammond: I'd rather we not support both versions. I don't want to accumulate version-specific cruft in a 0.X addon.

lib/engine-addon.js
```diff
@@ -387,10 +391,76 @@ module.exports = {
...
+          primaryStyleTree = concat(processedEngineStylesTree, {
+            allowNone: true,
+            inputFiles: ['**/*.css'],
+            outputFile: engineStylesOutputDir + 'engine.css'
+          });
+        }
+
+        var concatVendorCSSTree = concat(vendorCSSTree, {
+          allowNone: true,
+          inputFiles: ['**/*.css'],
```
nathanhammond: 😢

lib/engine-addon.js
rwjblue: Seems fine to me, I'll add an `ember-cli-version-checker` `assertAbove` for it.

lib/engine-addon.js
rwjblue: Added that fixture and rebased, https://travis-ci.org/ember-engines/ember-engines/jobs/192898083 shows no diff before and after this change.  

Also, if you are curious, the assets do seem to be appended to the correct place `dist/lazy-in-lazy/assets/engine-vendor.css`.

lib/engine-addon.js
rwjblue: Yep, good point.  Updating with comments...

lib/engine-addon.js
rwjblue: I don't really think this changes much, I'd prefer to duplicate than pull in a dep and join.

lib/engine-addon.js
rwjblue: 😿

lib/engine-addon.js
nathanhammond: Thank you for the new test case; this was my main concern I thought of with this change (though not necessarily stemming from this change).

lib/engine-addon.js
nathanhammond: Note this should probably mean that 0.5.0 is blocked on at least 2.12.0-beta.1.

rwjblue:
Thank you @trentmwillis, @nathanhammond, and @stefanpenner for reviewing and providing feedback.  I _think_ I have addressed all of the concerns at this point, so I am going to go ahead and land this.  If I've missed something please open an issue and I'll jump on it...

lib/engine-addon.js
rwjblue: Yep, agreed.  For now, I added a simple feature detect that throws an error if what we need is missing.  I totally agree that we shouldn't release 0.5.0 final until after there is a tagged/released build of ember-cli@2.12 (possibly even until 2.12.0 final is shipped).

PR #325, opened 2017-01-17
Update minimum version of broccoli-dependency-funnel.

rwjblue:


rwjblue:
Failures due to ember-cli version mismatch (which is being addressed in #324).

PR #330, opened 2017-02-02
Create .gitignore file in in repo addon blueprint

kiwiupover:
we should not check in `node_modules` pit of success for the win!


kiwiupover:
not needed in an in-app-addon

PR #331, opened 2017-02-02
Updated dependancies to show version of EmberCLI

kiwiupover:
I was confused a little by the talking just about Ember. Is there also a dependency on a version of emberJS?


knownasilya:
I think the above were referring to Ember, not ember-cli. But v0.5 does depend on latest alpha/beta of the cli I think.

kiwiupover:
Thanks @knownasilya I will update

PR #336, opened 2017-02-10
Ensure app/styles in child addons of lazy engines functions properly.

rwjblue:
Previously, if a lazy engine includes `app/styles` or any of its child addons include `app/styles` those styles were all being hoisted to the top level project's CSS assets.

This change handles the `styles` tree by ensuring that when the consuming app asks for `styles` tree we do not return one **AND** ensures that we do include `app/styles` in our custom build pipeline.

https://github.com/ember-engines/engine-testing/pull/12 was used to test this...

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -218,7 +218,17 @@ var buildCompleteJSTree = memoize(function buildCompleteJSTree() {
 });
 
 var buildEngineStyleTree = memoize(function buildEngineStyleTree() {
-  return this._treeFor('addon-styles');
+  // get engines own addon-styles tree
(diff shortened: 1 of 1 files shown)


rwjblue:
I reran the build and it seemed to work this time:

```
Executing: git clone --depth 1 https://github.com/ember-engines/engine-testing.git
Executing: npm link
Executing: npm install
Executing: bower install
Executing: ember build --output-path upstream-build
Executing: npm link ember-engines
Executing: ember build --output-path local-build
Executing: git diff --no-index upstream-build/ local-build/
ok 1 Acceptance build correctly builds engine-testing application
```

Issue #232, opened 2016-10-25
Lazy Loading Nonrouteable Engines


ooluwadiya:
I don't think this feature has been implemented / tested but I'll go ahead and log an issue for tracking. Lazy-loading non-routeable engines is not working. The assets are never loaded and as such are not registered...

I end up with this error  "Assertion Failed: You used `{{mount 'nonrouteable'}}`, but the engine 'nonrouteable' can not be found.""

trentmwillis:
Yeah, this hasn't really been considered yet. I think the primary question was how to handle the fact that the mount point will render asynchronously. Once that is determined, we just need to identify where we in that lifecycle we initiate the request to load the Engine assets.

ooluwadiya:
That makes sense. Thanks

aalasolutions:
Any updates on this? Our app is route-less and need to use lazy loading for the engine. Main JS file is like 3MB and some users are not having access to all the areas.

villander:
> Route-less Engines, on the other hand, are Engines which are not able to be routed to. This means that they don’t live at a specific route, but are instead mounted within a template or your application. Examples of good candidates for route-less Engines are chat boxes or complex forms.

Following the concept of route-less, I think we should start the request to load the Engine assets, when the template of that route / component is called. We can do something like what React does:

https://gist.github.com/rubenspgcavalcante/ea8d087cc2d997392abb2f1c7e02b666#file-app-jsx

> Yeah, this hasn't really been considered yet. I think the primary question was how to handle the fact that the mount point will render asynchronously. Once that is determined, we just need to identify where we in that lifecycle we initiate the request to load the Engine assets.

GCheung55:
The recent talk by [Dan Ambramov about async rendering](https://www.youtube.com/watch?v=v6iR3Zk4oDY) made me think of this lazy loading routless engines issue.

anulman:
Bumping this thread; per @villander's React notes I'd also suggest looking upstream at Webpack's code splitting guide, particularly re [prefetch & preload](https://webpack.js.org/guides/code-splitting/#prefetching-preloading-modules).

Since Ember is convention > configuration though, I think our syntax shifts slightly. Specifically, I think non-routable `{{mount 'foo'}}`s ought to:

- Accept an `eager=true` arg as a local override to the application's `lazyLoading: true` config—at build time this engine would be compiled into its host;
- Accept `prefetch=true` to lazy-load the engine once the host has completed loading & booting; and
[...]

Issue #117, opened 2016-05-17
Route less engine mounting with a variable


ssutar:
`{{mount someVar}}` does not work where `someVar` is a variable with the engine-name

It throws below error:
`Uncaught Error: Assertion Failed: The first argument of {{mount}} must be quoted, e.g. {{mount "chat-engine"}}.`

Not sure if this constraint is by design or an issue

mike183:
Looking at [the code](https://github.com/dgeb/ember-engines/blob/6c330b117b60de4f0cb4f88f3cfba7be0d6d2617/addon/-private/keywords/mount.js#L70-L73), it appears that this behaviour is intentional although I do agree that I think we should be able to pass a variable as the engine name, as long as the value of that variable was a string of course.

eccegordo:
Agreed. I am trying to make a component that dynamically mounts an engine. It would be nice if this was a little more flexible.

What about alternate syntax?

```
{{mountProp someProp}}
```

Also can someone explain why this doesn't work?

```
{{mount (concat "my" "-" "engine")}}
```

mike183:
Just submitted emberjs/ember.js#15015 which will enable routeless engines to be mounted using a bound property.

PR #338, opened 2017-02-14
[WIP] Refactor {{mount}} syntax

mike183:
Refactoring the `{{mount}}` syntax in order to lay the ground work required for #117 and #232.

cc/ @rwjblue

### DIFF
--- a/addon/-private/syntax-mount-ext.js
+++ b/addon/-private/syntax-mount-ext.js
@@ -1,124 +1,139 @@
 import Ember from 'ember';
 import emberRequire from './ext-require';
 
-const {
-  assert,
-  runInDebug
-} = Ember;
+const GlimmerRuntime = Ember.__loader.registry['@glimmer/runtime'];
+const GlimmerReference = Ember.__loader.registry['@glimmer/reference'];
 
-const mountSyntaxModule = Ember.__loader.require('ember-glimmer/syntax/mount');
-const ComponentDefinition = emberRequire('@glimmer/runtime', 'ComponentDefinition');
-const UNDEFINED_REFERENCE = emberRequire('@glimmer/reference', 'UNDEFINED_REFERENCE');
-const RootReference = emberRequire('ember-glimmer/utils/references', 'RootReference');
-const generateControllerFactory = emberRequire('ember-routing', 'generateControllerFactory');
-const OutletLayoutCompiler = emberRequire('ember-glimmer/syntax/outlet', 'OutletLayoutCompiler');
-const FACTORY_FOR = emberRequire('container', 'FACTORY_FOR');
-const AbstractManager = emberRequire('ember-glimmer/syntax/abstract-manager');
+if (GlimmerRuntime !== undefined || GlimmerReference !== undefined) {
 
-function dynamicEngineFor(vm, symbolTable) {
-  let env     = vm.env;
-  let args    = vm.getArgs();
-  let nameRef = args.positional.at(0);
+  const {
+    assert,
+    runInDebug
+  } = Ember;
 
-  return new DynamicEngineReference({ nameRef, env, symbolTable });
-}
+  const mountSyntaxModule = Ember.__loader.require('ember-glimmer/syntax/mount');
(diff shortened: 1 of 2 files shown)


addon/-private/syntax-mount-ext.js
```diff
@@ -0,0 +1,124 @@
...
+mountSyntaxModule.mountMacro = function(path, params, hash, builder) {
+  assert(
+    'You can only pass a single argument to the {{mount}} helper, e.g. {{mount "chat-engine"}}.',
+    params.length === 1 && hash === null
+  );
+
+  assert(
+    'The first argument of {{mount}} must be quoted, e.g. {{mount "chat-engine"}}.',
+    typeof params[0] === 'string'
+  );
```
chancancode: Wouldn't this defeat the purpose of this feature?

addon/-private/syntax-mount-ext.js
```diff
@@ -0,0 +1,124 @@
...
+    let { env, nameRef, /*symbolTable*/ } = this;
+    let nameOrDef = nameRef.value();
+
+    assert(
+      `You used \`{{mount '${nameOrDef}'}}\`, but the engine '${nameOrDef}' can not be found.`,
+      env.owner.hasRegistration(`engine:${nameOrDef}`)
+    );
+
+    return new MountDefinition(nameOrDef, env);
+  }
```
chancancode: We need to cache the definition here. If the returned value does not `===` the previously returned value, it will cause the component (the `{{mount`) to be torn down.

addon/-private/syntax-mount-ext.js
```diff
@@ -0,0 +1,124 @@
...
+
+  get() {
+    return UNDEFINED_REFERENCE;
+  }
```
chancancode: It is not necessary to have `get` here ✂️

addon/-private/syntax-mount-ext.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+class MountManager extends AbstractManager {
+  prepareArgs(definition, args) {
+    return args;
+  }
+
+  create(environment, { name, env }, args, dynamicScope) {
```
chancancode: `environment` === `env`, so we don't really need to store/destructure the `env` on the engine definition object

addon/-private/syntax-mount-ext.js
chancancode: I don't know what is the general rule.. but we should consider returning `null` here for production build (where the `assert` would have been stripped out).

addon/-private/syntax-mount-ext.js
```diff
@@ -0,0 +1,124 @@
...
+    runInDebug(() => this._pushEngineToDebugStack(`engine:${name}`, env));
+
+    dynamicScope.outletState = UNDEFINED_REFERENCE;
+
+    let engine = env.owner.buildChildEngineInstance(name);
+
+    engine.boot();
+
+    return { engine };
```
chancancode: We can just `return engine` here (and drop the destructuring in the other spots)

addon/-private/syntax-mount-ext.js
mike183: This PR is simply a refactor of existing functionality. The additional functionality will be added in follow up PRs.

addon/-private/syntax-mount-ext.js
mike183: Resolved.

addon/-private/syntax-mount-ext.js
mike183: ✂️

addon/-private/syntax-mount-ext.js
mike183: Resolved

addon/-private/syntax-mount-ext.js
mike183: Resolved

mike183:
I think I've covered all of the concerns that were raised. Tests are currently failing but the reason seems unrelated as the same failure also appears to be occurring on master.

mike183:
This PR has been superseded by https://github.com/emberjs/ember.js/pull/14967

PR #342, opened 2017-02-21
Fix the build.

rwjblue:
Due to the issue reported in https://github.com/ember-cli/ember-cli/issues/6779 our CI was passing when tests were actually failing. This PR is addressing the underlying reason they were failing (as https://github.com/ember-cli/ember-cli/issues/6779 will be fixed upstream via changes in ember-cli itself).

* Remove unneeded shimming around for `modules/` paths (this was not needed in ember-cli@2.12 and actually results in no file matches in ember-cli@2.13+).
* Fix test assertions that were confirming number of asset files loaded. The changes in https://github.com/ember-engines/ember-engines/pull/336 lead to us having an actual CSS file asset also, and the tests reported `not ok` but CI still passed (so we merged the PR). This fixes the assertions, and updates to have the tests actually pass for realz.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -113,13 +113,7 @@ var buildVendorJSTree = memoize(function buildVendorJSTree(vendorTree) {
     exclude: ['vendor/**/*.*']
   });
 
-  var vendorJSTreeRelocated = new Funnel(vendorJSTree, {
-    srcDir: 'modules',
-    destDir: '/',
-    allowEmpty: true
-  });
-
-  return vendorJSTreeRelocated;
+  return vendorJSTree;
 });
 
 var buildVendorCSSTree = memoize(function buildVendorCSSTree(vendorTree) {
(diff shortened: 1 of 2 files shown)


rwjblue:
FYI - The `node:ci` build is failing (correctly) because the before and after diff is not the same.  Prior to these changes (when using ember-cli#master) the `dist/engines-dist/ember-blog/engine.js` file was completely empty, after these changes it properly has content.

rwjblue:
Confirmed that `ember test` passes when using either ember-cli@2.12.0-beta.1 *or* ember-cli#master.

rwjblue:
> This means that 0.5 will only be compatible with ember-cli@^2.12.0 correct?

Yes, that is already the case though (as we assume things like `this._addonPreprocessTree` exist which are only present in 2.12).  It is likely possible to make compatible with 2.11 by adding some additional feature detect code, but I'm on the fence if we should. Happy to hash that out separately before releasing 0.5.0 stable though...

PR #345, opened 2017-02-24
Ensure dependency styles are in same relative path as engine styles.

rwjblue:
When we call `treeFor('addon-styles')` the tree is returned with the `addon/styles` contents in `/`, however calling `this.eachAddonInvoke('treeFor', ['styles'])` returns the contents of dependent addons' `app/styles` tree _inside_ `app/styles`.

This change ensures that our dependencies `app/styles` contents are at the same level as our own `addon/styles` (and therefore can be imported from).

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -426,7 +426,7 @@ module.exports = {
         if (engineStylesTree) {
           var preprocessedEngineStylesTree = this._addonPreprocessTree('css', engineStylesTree);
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Reproduction of the original issue in https://github.com/ember-engines/engine-testing/pull/13.

rwjblue:
>  Should we try and apply maybeMergeTrees elsewhere in a follow up?

Yes, that was exactly my thought.  It should be a nice (but tiny) boost, and likely clean up a few gnarly merges...

PR #353, opened 2017-03-03
Avoid eagerly populating QP meta when possible.

rwjblue:
Upstream changes are removing the `_populateQPMeta` method on `Ember.Route` instances (in order to prevent the corresponding controller from being eagerly looked up).

This preserves the same semantics, but avoids usage of `route-meta` and invocation of `handler._populageQPMeta` if it is not present.


rwjblue:
No, it should not (we still call `handler._populateQPMeta()` if present) affect version compat.

Issue #329, opened 2017-02-01
Deduplicate Addon Assets


trentmwillis:
Currently, when an addon is included in multiple locations with lazy Engines, that addon's code is duplicated. This problem breaks down into two scenarios:

1. Duplication when **a parent and child** have the same addon. This is easy to solve, as the parent copy should be the only copy, since it is guaranteed to be loaded first.

2. Duplication when **siblings** have the same addon. This is difficult to solve as we can't make assumptions as to which Engine will be loaded first. There are three potential approaches in this scenario:

    * Include the asset in both Engine bundles -> this is easy to implement and use (and is our current reality), but has negative impact on asset download size for our users. Also has potential to run into weird, state issues by clobbering previously loaded copies of the module.
    * Bundle shared assets separately and load with first Engine that needs them -> relatively easy to implement and use, but can incur multiple network requests for potentially small assets. The Asset Manifest/Loader has easy support for this use case.
    * Hoist shared assets into the closest common ancestor's bundle -> no more difficult to implement than the above, but bloats the main assets with potentially unused features. Also can make it difficult to reason about where assets wind up.

It is likely that each of the above approaches can be good or bad depending on circumstances.

Thus, we need to discuss and arrive at a solution for deduplication in **sibling** engines. The **parent and child** scenario just needs to be implemented.

cibernox:
Point one is crystal clear, agree.

About point 2, I think that all three approaches are valid options depending on your app. I don't think there is a single solution. 'd like the decission to be in the hands of the developer. Ideally expressed just in terms of package.json, but I don't know if it will be doable.

By example, the last option, hoist shared addons to the parent app is as simple for the user as include that addon in the package.json of the parent app. That would mean we're now in point 1, in which we agree there it should be hoisted.

However repeat the addon or bundle it as a separate file is tricky and will probably depend on your app and the addon. An entra request might make sense for big addons, not so much for those that just give you a couple helpers. It will become more and more if your app leverages SW download deltas between versions.

Without giving it too much though I think that the default approach for duplicated addons that are not in the parent app could stay as today and duplicate them, and if the user wants to bundle some addons separately those can be listed in some sort of special config key, either in the package.json or in the ember-cli-build.js of the host app.

trentmwillis:
A PR has been opened for Point 1: https://github.com/ember-engines/ember-engines/pull/356.

For Point 2, I think the approach @cibernox outlines is good. Which means for the 3 scenarios I outline above we have:

[...]

PR #356, opened 2017-03-13
Deduplicate addons that already appear in a host

trentmwillis:
Addressing the first part of https://github.com/ember-engines/ember-engines/issues/329.

Figuring out duplicated addons is a bit more tricky than I anticipated. Largely because `eachAddonInvoke` is essentially a depth-first walk of the addon tree, but we would want it to be breadth-first so that we know all the addons at the previous level.

Note: I added an explicit CSS file to the ember-blog engine, because the styles tree would be undefined now that we deduplicate the styles trees.

Todo:
- [x] Tests

### DIFF
--- a/node-tests/acceptance/build-test.js
+++ b/node-tests/acceptance/build-test.js
@@ -10,20 +10,10 @@ const build = require('../helpers/build');
 const InRepoAddon = require('../helpers/in-repo-addon');
 const InRepoEngine = require('../helpers/in-repo-engine');
 const expectedManifests = require('../fixtures/expected-manifests');
+const matchers = require('../helpers/matchers');
 
-/**
- * Creates a regex that matches the definition for the specified module name.
- *
- * @param {String} moduleName
- * @return {RegExp}
- */
-function moduleMatcher(moduleName) {
-  return new RegExp(`define\\(['"]${moduleName}['"]`);
-}
-
-function cssCommentMatcher(comment) {
-  return new RegExp(`/\\* ${comment} \\*/`);
-}
+const moduleMatcher = matchers.module;
+const cssCommentMatcher = matchers.cssComment;
 
 describe('Acceptance', function() {
   describe('build', function() {
--- a/node-tests/acceptance/build-test.js
+++ b/node-tests/acceptance/build-test.js
@@ -7,6 +7,7 @@ const expect = require('chai').expect;
 const AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
 
 const build = require('../helpers/build');
+const InRepoAddon = require('../helpers/in-repo-addon');
 const InRepoEngine = require('../helpers/in-repo-engine');
 const expectedManifests = require('../fixtures/expected-manifests');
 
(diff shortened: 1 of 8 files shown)


trentmwillis:
@rwjblue could the deduplicating be causing one of the CSS files to not be generated now because it would be empty? Trying to determine if the test failures are legit or not

trentmwillis:
Updated and figured out that this isn't really a change in behavior. It's just that since we deduplicated the styles tree, that tree is now totally empty instead of previously have some empty trees in it.

trentmwillis:
@rwjblue updated with further tests. One of them, however, is currently failing:

> * Shared addon deeper than engine:
>   * project has dep on addon-derp
>   * project has lazy engine engine-a
>   * engine-a has engine addon-z
>   * addon-z has addon-derp (addon-derp should not be duplicated)

This doesn't currently work because we cannot `nonDuplicateAddonInvoke` from within an Engine. Not sure it'll be possible to properly handle this without a lot of monkey-patching or upstream changes.

rwjblue:
Ya, that was the scenario that I thought would not pass. Seems good to land this though, as it is still a good improvement on what we have today. We can always figure out how to make it better...

node-tests/acceptance/dedupe-addons-test.js
```diff
@@ -0,0 +1,260 @@
...
+const InRepoAddon = require('../helpers/in-repo-addon');
+const InRepoEngine = require('../helpers/in-repo-engine');
+
+/**
+ * Creates a regex that matches the definition for the specified module name.
+ *
+ * @param {String} moduleName
+ * @return {RegExp}
+ */
+function moduleMatcher(moduleName) {
```
rwjblue: Can we share this util instead of duplicating?

node-tests/acceptance/dedupe-addons-test.js
```diff
@@ -0,0 +1,260 @@
...
+  return new RegExp(`define\\(['"]${moduleName}['"]`);
+}
+
+function cssCommentMatcher(comment) {
```
rwjblue: Share?

Cryrivers:
I would like to add one more test case, which is for `ember-composable-helpers`

say host and engine are both using `ember-composable-helpers` but with different settings. you could check out README.md of https://github.com/Cryrivers/ember-engines-dedupe-vendor

trentmwillis:
Thanks that use case @Cryrivers. @rwjblue how do you think we should handle that? I kind of think we should punt on it for now and just force the host to include a superset of the includes; primarily because to support this properly would rely on designing an API to give addons more control.

rwjblue:
>  primarily because to support this properly would rely on designing an API to give addons more control

I think we already have this mechanism, it was added via https://github.com/ember-cli/rfcs/pull/91.  We are no-oping specific trees based on duplication between parent and lazy child. This isn't terribly different from what we did in ember-cli around the deduplication of invoking these hooks. We could use the cacheKeyForTree from the parent's version of the addon for a given tree to determine if the lazy child version should be no-op'ed.

trentmwillis:
So basically, if we can't cache a tree from an addon, then we can't safely dedupe it? Seems fine to me.

rwjblue:
> I kind of think we should punt on it for now and just force the host to include a superset of the includes

I'm totally fine with landing this PR as is, and iterating forward to rely on `cacheKeyForTree`...

rwjblue:
> So basically, if we can't cache a tree from an addon, then we can't safely dedupe it?

Yeah, to me it seems like basically the same fundamental question: 

> hey, addon-xyz, are you doing something different in `treeFor('blah')` between these two instances?

trentmwillis:
Yep yep, makes sense. I'll update to address the current comments and we can follow up on @Cryrivers use case in another PR.

node-tests/acceptance/build-test.js
```diff
@@ -10,20 +10,12 @@ const build = require('../helpers/build');
...
-  return new RegExp(`define\\(['"]${moduleName}['"]`);
-}
-
-function cssCommentMatcher(comment) {
-  return new RegExp(`/\\* ${comment} \\*/`);
-}
+const {
+  module: moduleMatcher,
+  cssComment: cssCommentMatcher
+} = matchers;
```
rwjblue: Destructuring is not supported in node 4. I'm unsure if we run our test suite there though...

trentmwillis: So fast at reviewing...yeah, I was unsure if it was supported or not. We do test in Node 4. I'll change this.

PR #357, opened 2017-03-14
Make the default addon tree an option

### LINKED ISSUE #355: treePath is hardcoded to a path instead of using the originally defined treeForAddon
[engine-addon.js#L129](/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L129) is using a hardcoded path `var treePath = path.resolve(this.root, this.treePaths['addon']);` instead of using the output of the originally defined `treeForAddon` function. This makes it difficult for an engine to modify it's build.

I spoke with @rwjblue about this on slack and he confirmed it is a bug.

jsturgis:
This change fixes an issue where the treePath is hardcoded to a path instead of being configurable by the user. 

Fixes #355

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -26,6 +26,9 @@ var DEFAULT_CONFIG = {
       css: '/assets/engine-vendor.css',
       js: '/assets/engine-vendor.js'
     }
+  },
+  trees: {
+    addon: 'addon'
   }
 };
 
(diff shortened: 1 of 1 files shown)


lib/engine-addon.js
```diff
@@ -126,9 +129,17 @@ var buildVendorCSSTree = memoize(function buildVendorCSSTree(vendorTree) {
 
 var buildEngineJSTree = memoize(function buildEngineJSTree() {
   var engineSourceTree;
-  var treePath = path.resolve(this.root, this.treePaths['addon']);
-  if (existsSync(treePath)) {
+  var treePath;
+  var addonTree = this.options.trees.addon;
```
trentmwillis: Is it not possible for us to just override `treePaths` directly via a merge or something?

rwjblue: The key is that we expect a tree not a path.  `treePaths` is for how we represent "where" things are, but this is specifically to allow custom filtering and whatnot of the tree.

trentmwillis: Ah, okay I see. Since a string value is the default here I didn't consider setting it to a non-string value.

jsturgis:
I rebased off master.

PR #360, opened 2017-03-15
Implement better testing story for build

trentmwillis:
The current testing setup is super opaque and brittle. This PR explores a setup to allow us to test different combinations of engines/addons in an application build within our repo. This should hopefully make the tests less brittle while giving us better insight into what parts of the system are under test.

TODO:
- [x] Cleanup assertions
- [x] Update asset-manifest.json assertions to verify contents

*Note: Apparently route hoisting is broken on master right now. Will need to fix that.


node-tests/acceptance/build-test.js
```diff
@@ -1,31 +1,349 @@
...
+    }));
+
+    it('invokes trees and executes hooks in the proper order and handles imports for engines', co.wrap(function* () {
+      let app = new AddonTestApp();
+      let engineName = 'tree-invocation-order';
+
+      yield app.create('engine-testing', { noFixtures: true });
+      let engine = yield InRepoEngine.generate(app, engineName, { lazy: true });
+
+      let fixture = fixturify.readSync(path.resolve(__dirname, '../fixtures/tree-invocation-order'));
```
rwjblue: I'd prefer to store this in a single file, instead of using `fixturify.readSync`.  I find that it makes it much much easier to grok if I can see the structure in one place.

The file can look like:

```js
const stripIndent = require('common-tags').stripIndent;

module.exports = {
  'some/path': stripIndent`
    whatever text contents go here
  `
}
```

Then just pass that in to `engine.writeFixture`

node-tests/acceptance/build-test.js
```diff
@@ -1,31 +1,349 @@
...
+    }));
+
+    it('correctly hoists routes.js and its imports', co.wrap(function* () {
+      let app = new AddonTestApp();
+      let engineName = 'routes-hoisting';
+
+      yield app.create('engine-testing', { noFixtures: true });
+      let engine = yield InRepoEngine.generate(app, engineName, { lazy: true });
+
+      let fixture = fixturify.readSync(path.resolve(__dirname, `../fixtures/${engineName}`));
```
rwjblue: Thoughts on suggestion above?

rwjblue:
To confirm:

> *Note: Apparently route hoisting is broken on master right now. Will need to fix that.

this is fixed now, right?

trentmwillis:
> How slow is the suite ATM?

Only slightly slower than the `try:each`, so around 5-6 minutes. Most of that is spent on the first test due to npm installation.

And yes, route hoisting is fixed now.

node-tests/acceptance/build-test.js
trentmwillis: Sounds good to me! I set it up this way initially because I basically just copy-pasta'd from the engine-testing repo.

rwjblue:
Awesome. Can we use `yarn` for the initial install?

trentmwillis:
> Need to add more assertions around CSS file contents

Can you elaborate more on what exactly should be tested? I know we need coverage, I'm just not sure what it is as I'm still pretty unfamiliar with that space.

> Awesome. Can we use `yarn` for the initial install?

Yes. I'd like to switch to `yarn`, but think we should do that in a follow up to avoid this getting too complicated.

node-tests/acceptance/build-test.js
rwjblue: ya, makes sense.  And we still may want separate file for [index.js](https://github.com/trentmwillis/ember-engines/blob/173c5969d772817adef61ca508a81b41cba833a4/node-tests/fixtures/tree-invocation-order/index.js), but most of the other files would easily be inlined.

Totally up to you on inlining all or just the "smallish" ones...

rwjblue:
> Yes. I'd like to switch to yarn, but think we should do that in a follow up to avoid this getting too complicated.


SG, I'm happy to do that in a separate PR, or we can wait for this to land (to avoid rebase hell). Do you have a preference?

trentmwillis:
I'd prefer to do it after this lands. I want to update ember-cli-addon-tests to use `yarn` if possible and use that with this as well. Hoping that will save a couple minutes.

PR #368, opened 2017-03-20
Remove Ember Try scenario for Ember 1.13

mike183:
AFAIK Ember 1.13 isn't supported in any version of ember-engines.


rwjblue:
We should probably also add running ember-try to the travis config.

rwjblue:
Thanks for updating, we can do a follow up for adding this to Travis.

PR #370, opened 2017-03-20
Fix deprecation warnings

mike183:
Fixes some minor deprecation warnings that were being thrown when running the tests.


rwjblue:
Thank you!

PR #374, opened 2017-03-20
Remove bower (mostly)

trentmwillis:
🔥 🔪 🔥


rwjblue:
Looks like errors thrown in ember-cli-addon-tests when bower is missing...

trentmwillis:
Yeah, looks like it is going to be annoying to unwind, but going to work on it

trentmwillis:
I updated this to mostly remove Bower. We still install it globally during CI so that ember-cli-addon-tests can access it when creating a new app.

PR #384, opened 2017-03-24
Ensure `asset-loader` runs after `broccoli-asset-rev`

spencer516:
This ensures that the asset manifest generated by the `asset-loader` will be fingerprinted by `broccoli-asset-rev`. Which, in turn, ensures that the `{app}/config/asset-manifest` meta tag's value will contain the correct fingerprinted assets. 

(I was able to successfully deploy an app with fingerprinting + async engines after applying this fix).

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "lodash.defaultsdeep": "^4.5.1"
   },
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "after": "broccoli-asset-rev"
   }
 }


rwjblue:
Just to clarify, we want the [postprocessTree](https://github.com/ember-engines/ember-asset-loader/blob/master/lib/manifest-generator.js#L33) from `ember-asset-loader`'s `ManifestGenerator` (used by ember-engines's itself [here](https://github.com/ember-engines/ember-engines/blob/master/index.js#L6)) to be ran before `broccoli-asset-rev`'s own `postprocessTree` so that the filenames listed in `dist/asset-manifest.json` have been fingreprinted, right?

I'm unsure that this PR actually does that. I would have thought that we actually need to specify _before_ not _after_ for this...

dgeb:
Closed via #632

Issue #310, opened 2017-01-11
Trim down README


rwjblue:
Now that we have a nice website, we should trim down the README a bit I believe. The README should end up being a funnel to either ember-engines.com or `CONTRIBUTING.md`.

samselikoff:
Went through the readme and site, here's the TODO list. Some questions I have for yall here.

- [ ] 1. The README has this:

    > The following functionality will soon be supported:
    > - Route serializer modules that isolate serialization logic from the rest of the route definition.

    What does this mean?

- [ ] 2. _"Namespaced access to engine resources from applications."_ Is this still under consideration? Should it be on the website?

- [ ] 3. Does the _Linking and external links_ section of the website need to mention the [Engine Linking RFC](https://github.com/emberjs/rfcs/pull/122)?

- [ ] 4. Note from Dependencies section: _"Also note that multiple engines can be configured per parent application/engine, and that each engine name should be camelCased (emberBlog instead of ember-blog)."_ Should this be moved over?

- [ ] 5. Testing - what's the story here? Website says PRs welcome, README has some notes on testing components within an in-repo engine. I guess we could migrate this, and write a note about lack of a story for acceptance-level tests and testing lazy engines?

- [ ] 6. Need to add notes about versions/compatibility

- [ ] 7. Add link to Dan's video somewhere (Additional Resources probably)

- [ ] 8. Looks like the entire [Accessing Engine configuration settings](https://github.com/ember-engines/ember-engines#accessing-engine-configuration-settings) section needs to be added, I will look into this

- [ ] 9. Details on build output. Website currently says _"For more details on the build output and what each file represents, check out the Built Engine Output section of the ember-engines readme."_ How much of this do we want on the website, and do we want a new section?

- [ ] 10. Should we rename "Mounting an engine" section of website to "Consuming an engine"? Add note about `ember install your-engine`.

- [ ] 11. Link to dgeb's demo apps

trentmwillis:
@samselikoff sorry for letting this sit for so long. I've responded to all your points below. I'd appreciate it if you could work on some of these, but I'll also plan to tackle some of them.

1. Refers to the [Route Serializers RFC](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md), which is mentioned in guides [here](http://ember-engines.com/guide/lazy-loading).

2. I'll defer to @rwjblue, but I think we should drop it for now.

3. Yes, it should.

4. Yes, it should as I don't believe it is mentioned.

[...]

PR #387, opened 2017-04-10
Removed content no longer  needed because of ember-in-repo-engine

perlun:
These files are already created with the correct content by the blueprint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,59 +86,6 @@ ember g in-repo-engine <engine-name>
 Don't forget to install `ember-engines` and the appropriate version of Ember in
 your project, as described above.
 
-### Configuring your Engine
-
-Within your engine's root directory, modify `index.js` so that your addon
-is configured as an engine using the `EngineAddon` extension:
-
-```js
-/*jshint node:true*/
-const EngineAddon = require('ember-engines/lib/engine-addon');
-
-module.exports = EngineAddon.extend({
-  name: 'ember-blog'
-});
-```
-
-Within your engine's `config` directory, create a new `environment.js` file:
-
-```js
-/*jshint node:true*/
-'use strict';
-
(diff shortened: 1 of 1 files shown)


mike183:
Wouldn't these steps still be required when creating a standalone engine? Which doesn't yet have a blueprint.

perlun:
> Wouldn't these steps still be required when creating a standalone engine? Which doesn't yet have a blueprint.

Ah, you might be right. Didn't think about that. 😏 Should we then just change this to be a note saying "you don't need this if you're creating an in-repo-engine" or something to the like?

trentmwillis:
We actually need to cut out most of the stuff from the readme, there is an open issue (which I have meant to get to, but no time): https://github.com/ember-engines/ember-engines/issues/310

mike183:
@trentmwillis Is there much stopping us creating a blueprint for standalone engines?

I know there were previously reasons we didn't have one, but unsure what they were.

trentmwillis:
@mike183 the previous issue was there is no way to install a blueprint globally, thus you'd have to be in an Ember project before being able to use the Engine blueprint. That should be unblocked with the next release of Ember-CLI since it'll allow you to download blueprints (such as for glimmer.js apps). @rwjblue could chime in if there is anything else needed.

rwjblue:
Nope, thats it, we could even start the blueprint now (and it would only be usable with canary until the next beta branch point)...

rwjblue:
There is now a blueprint for stand alone engines (as well as one for in-repo engines).

@mike183 - Can you review this again?

perlun:
@trentmwillis 

> These changes seem good to me. Thanks! And sorry for this taking so long to merge.

No worries! There were good reasons for the delay in this case. 👍

PR #390, opened 2017-04-12
Add deprecated flag to force ember-engines@0.4 styles behaviors.

rwjblue:
ember-engines@0.4 had a number of flaws with its style processing. A few examples of this are:

* All engine dependencies that have `app/styles` trees were **always** hoisted to the top level host application.
* Preprocessors were being ran only for the engines own styles, but not including any of its dependencies `app/styles` (causing `@import` to work in very bizarre ways).
* `preprocessTree` / `postprocessTree` hooks were never called for styles

All of these issues have been addressed in ember-engines@0.5, and so far there are no known issues with the fixes added to 0.5 for this. Unfortunately, applications in the wild came to rely on the broken behaviors :cry:, and now have completely broke styles when upgrading to ember-engines@0.5. 

This PR adds a mechanism for applications to leverage which allows a per-engine opt-out of the new styles behaviors. When enabled, this flag causes the engine to behave nearly identically to ember-engines@0.4 (WRT styles).

Usage of this flag would look like:

```js
const EngineAddon = require('ember-engines/lib/engine-addon');

  name: 'ember-blog',
  useDeprecatedIncorrectCSSProcessing: true
});
```

*Note:* Enabling this flag, will emit a console warning (for each engine that it is enabled on).  This backwards compatibility shim will be removed prior to ember-engines@0.6.

---

For ease in reviewing, the primary changes from 0.4 to 0.5 on the styles front were:

* https://github.com/ember-engines/ember-engines/pull/324
* https://github.com/ember-engines/ember-engines/pull/336
* https://github.com/ember-engines/ember-engines/pull/345

This PR is intending to nullify all three of those PR's when the flag is enabled...

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -858,7 +858,9 @@ module.exports = {
       var trees;
       if (
         (name === 'app') ||
-        (name === 'styles') ||
+          // when using deprecated 0.4 style processing, we allow the styles tree to be
+          // hoisted to the top level host
+        (this.useDeprecatedIncorrectCSSProcessing !== true && name === 'styles') ||
         (name === 'addon' && this.lazyLoading === true) ||
         (name === 'public' && this.lazyLoading === true)
       ) {
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -372,6 +372,11 @@ module.exports = {
 
       var result = originalInit.apply(this, arguments);
 
+      var useDeprecatedIncorrectCSSProcessing = this.useDeprecatedIncorrectCSSProcessing === true;
+      if (useDeprecatedIncorrectCSSProcessing) {
+        this.ui.writeDeprecateLine(this.pkg.name + ' engine has opted in to ember-engines@0.4 styles processing compatibility mode (via use of the `useDeprecatedIncorrectCSSProcessing` flag), this is deprecated and will be removed in future versions.');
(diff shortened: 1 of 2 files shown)


lib/engine-addon.js
```diff
@@ -372,6 +372,11 @@ module.exports = {
 
       var result = originalInit.apply(this, arguments);
 
+      var useDeprecatedIncorrectCSSProcessing = this.useDeprecatedIncorrectCSSProcessing === true;
+      if (useDeprecatedIncorrectCSSProcessing) {
+        this.ui.writeDeprecateLine(this.pkg.name + ' engine has opted in to ember-engines@0.4 styles processing compatibility mode, this is deprecated and will be removed in future versions.');
```
trentmwillis: Nit: Should note that "engine has opted in to ember-engines@0.4 styles processing compatibility **via use of the useDeprecatedIncorrectCSSProcessing flag**". Just to make it clear for those that might not be familiar with the option.

rwjblue: Good point, will updated

rwjblue:
Updates:

* Added tests with flag on and off (showing differing behaviors)
* Updated deprecation message to make it clear this is about `useDeprecatedIncorrectCSSProcessing` flag
* Removed `WIP` label, I believe this is ready to go

PR #393, opened 2017-04-17
Change useDeprecatedIncorrectCSSProcessing to emulate 0.5.0-beta.6.

rwjblue:
The flag previously emulated 0.4.0, this updates the flag to emulate 0.5.0-beta.6.

The primary difference here is that https://github.com/ember-engines/ember-engines/pull/324 is now included in these changes...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "4"
+  - "6"
 
 cache:
   yarn: true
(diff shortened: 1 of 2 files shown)


rwjblue:
UGH, CI is broken because the latest version of `yarn` does not support Node 4 (see https://github.com/yarnpkg/yarn/issues/3138 and https://github.com/yarnpkg/yarn/pull/3139).

mbutki:
I tested it and it works for our use case.

PR #402, opened 2017-05-02
Only include app/styles from dependency addons in lazy case

dnachev:
This is more like a conversation starter rather than a proper fix for a problem. The behavior I'm observing is that styles are duplicated in both engine.css and engine-vendor.css. I think this is caused, because engines, which are acting as addons (non-lazy engines), will build their own styles (everything in their addon/styles + dependency addon/styles + app/styles). When acting as lazy engines, however, it is presumed that the engine needs to build the app-namespaced styles as well (similar to what the application will do). However, they are already built as part of the addon styles and therefore duplicated.

Changing the code to only include the app/styles from the dependencies in case of a lazy engine, ensures they are not duplicated in both engine.css and engine-vendor.css.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -255,27 +255,6 @@ function maybeMergeTrees(_inputTrees, options) {
   }
 }
 
-var buildEngineStyleTree = memoize(function buildEngineStyleTree() {
-  // gather engines own `addon/styles` and its dependencies `app/styles`
-  var engineStylesTree = this._treeFor('addon-styles');
-  var dependencyStyleTrees = this.nonDuplicatedAddonInvoke('treeFor', ['styles']);
-  var dependencyStyleTree = maybeMergeTrees(dependencyStyleTrees, { overwrite: true });
-  var relocatedDependencyStyleTree;
-
-  // if dependency styles trees were found, relocate them to the expected
-  // path (`addon/styles)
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks for opening this for discussion! 

I think the best way to move forward here would be to add a specific test for this scenario.  

For both eager and lazy engine we should:

* add a dep of the engine with `app/styles`
* enable a preprocessor (ember-cli-sass most likely) in the engine
* `@import 'some-file-from-deps-app-styles'`
* ensure the contents are in the correct place
  * Lazy engines: contents should be in `engine.css` (and not `engine-vendor.css`)
  * Eager engines: contents should be in the host app's `vendor.css` (and not in either `engine.css` or `engine-vendor.css`)

dnachev:
OK, after chatting offline with @rwjblue the fix is not correct. The engine is expected to act as an application and therefore it has to include its dependencies app/styles, even if it is not lazy. The specific problem I was hitting was caused by duplicate dependencies in a parent engine (e.g. the same dependency was added twice). Because each engine is responsible to compile its CSS, duplicated dependencies will produce results in both engine-vendor.css (coming from the nested engine) and the engine.css (coming from the parent engine). The only way to avoid this is to ensure there is no duplication in the dependencies. If there is, there will be duplication of CSS as well, but semantically this is correct behavior.

Closing this.

Issue #265, opened 2016-11-18
Compact re-exports for Engines


trentmwillis:
Currently, when using the `EngineAddon`base class, we automatically re-export dependency's `app` folder contents in the Engine's namespace. This helps make it easy for consumers to use Addons in their Engines. However, the compiled code for this can be quite bloated:

```js
define('ember-chat/components/spanish-greeting', ['exports', 'common-components/components/spanish-greeting'], function (exports, _commonComponentsComponentsSpanishGreeting) {
  'use strict';

  Object.defineProperty(exports, 'default', {
    enumerable: true,
    get: function get() {
      return _commonComponentsComponentsSpanishGreeting['default'];
    }
  });
});
```

For smaller applications, this isn't that big of a deal, but for larger applications with many Engines and many Addons, this duplication can start to add up.

This can be worked around by modifying the resolver to check additional namespaces, but this is sub-optimal as it incurs a runtime penalty for each additional lookup.

What I want to suggest instead is that we leverage the aliasing capabilities of `loader.js` and do the following:

```js
define('ember-chat/components/spanish-greeting', define.alias('common-components/components/spanish-greeting');
```

As far I can tell, this approach is functionally equivalent since the exports should still be the same reference. We could potentially compact this even further and do something to the effect of:

```js
const reexports = [
  ['ember-chat/components/spanish-greeting', 'common-components/components/spanish-greeting'],
  // ...
].forEach(([alias, original]) => define(alias, define.alias(original)));
```

Thoughts?

cc @stefanpenner @nathanhammond @rwjblue

nathanhammond:
Things to think about:

- Does this approach break `instanceOf` checks?
- What do we do in a world where `loader.js` is optional? (AKA: the more static we make things, the less we need `loader.js`)

The "namespace fallback" resolver slows down _all_ lookups, but if there are fewer resolutions required due to things being more static it might still be a tenable solution. It is also guaranteed to be the most terse of the solutions as the module names needn't be enumerated.

(Stream of consciousness response.)

trentmwillis:
> Does this approach break `instanceOf` checks?

Unsure which specific checks you're referring to, but the reference returned by the module is the same. So objects will be the same as will classes, functions, etc.. This is the same as for traditional re-exports.

> What do we do in a world where `loader.js` is optional?

Is there actually a plan to drop `loader.js` any time in the near future?

> if there are fewer resolutions required due to things being more static it might still be a tenable solution.

[...]

PR #408, opened 2017-05-05
Support compact re-exports

trentmwillis:
Fixes https://github.com/ember-engines/ember-engines/issues/265. Leverages [babel-plugin-compact-reexports](https://github.com/ember-engines/babel-plugin-compact-reexports).

TODO:
- [x] Tests

FOLLOW UPS:
- [x] Add tests for test matcher utilities
- [x] Add eslint

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -13,6 +13,8 @@ var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 var Addon = require('ember-cli/lib/models/addon');
 var memoize = require('./utils/memoize');
 var p = require('ember-cli-preprocess-registry/preprocessors');
+var CompactReexports = require('babel-plugin-compact-reexports');
+var semver = require('semver');
 var preprocessCss = p.preprocessCss;
 var BroccoliDebug = require('broccoli-debug');
 
@@ -367,6 +369,22 @@ module.exports = {
 
     options.debugTree = BroccoliDebug.buildDebugCallback('ember-engines:' + options.name);
 
+    /**
+     * Determines if ember-engines should use the compact-reexports Babel plugin.
+     */
+    options.shouldCompactReExports = function() {
+      var VersionChecker = require('ember-cli-version-checker');
+      var checker = new VersionChecker(this);
+
+      var emberCliBabelVersion = this.addons.find(addon => addon.name === 'ember-cli-babel').pkg.version;
+
+      // We check the local ember-cli-babel version and the global ember-cli and
+      // loader.js versions
+      return semver.satisfies(emberCliBabelVersion, '>=6.0.0') &&
+        checker.for('ember-cli', 'npm').satisfies('>=2.13.0') &&
+        checker.for('loader.js', 'npm').satisfies('>=4.4.0');
+    };
+
     options.init = function() {
       this.options = defaultsDeep(options, DEFAULT_CONFIG);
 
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
   ],
   "dependencies": {
     "amd-name-resolver": "0.0.5",
+    "babel-plugin-compact-reexports": "^0.1.0",
     "broccoli-babel-transpiler": "^5.6.1",
```
rwjblue: We need to remove our direct usage of `broccoli-babel-transpiler` and instead swap to the new API in ember-cli-babel@6 that is documented [here](https://github.com/babel/ember-cli-babel#addon-usage). This also allows us to remove `DEFAULT_BABEL_OPTIONS`.

lib/engine-addon.js
```diff
@@ -362,6 +362,18 @@ module.exports = {
...
 
+    /**
+     * Determines if ember-engines should use the compact-reexports Babel plugin.
+     */
+    options.shouldCompactReExports = function() {
+      var VersionChecker = require('ember-cli-version-checker');
+      var checker = new VersionChecker(this);
+
+      return checker.for('ember-cli', 'npm').satisfies('>=2.13.0') &&
+        checker.for('ember-cli-babel', 'npm').satisfies('>=6.0.0') &&
```
rwjblue: This isn't quite right (sadly).  This will check if the top level project is using ember-cli-babel@6, but not if _this_ engine is (in its own `package.json`).  I'm working on refactoring ember-cli-version-checker to inherently support this as a concept (I need it for exactly this reason also), but it does not yet.

rwjblue: Checking for ember-cli and loader.js here is fine, because those are actually project level deps.

trentmwillis: @rwjblue in the interim would the best solution be to `findAddonByName` check its package directly? (Note: I could also work on adding this support to ember-cli-version-checker.)

rwjblue: I got lost down the rabbit hole in the version checker, I ended up completely refactoring it to use ES6 before working on the thing I actually needed to do.

> in the interim would the best solution be to findAddonByName check its package directly?

Yes, `this.addons.find(addon => addon.name === 'ember-cli-babel').pkg.version` or something?

package.json
trentmwillis: I don't think we can remove that as we directly invoke it in the case that we're not using Babel@6...unless we want to drop support for that use case. Regardless, does this need to be done in this PR or could it be a follow-up task?

trentmwillis:
Note: I've opened https://github.com/ember-engines/babel-plugin-compact-reexports/pull/4 to fix the caching strategy for the new Babel plugin.

package.json
rwjblue: Follow up, I just noticed during review.

trentmwillis:
@rwjblue I've added tests. So whenever you have a moment, I'd appreciate another review.

trentmwillis:
Updated again to add unit tests for the test helpers we use (such meta).

yarn-error.log
```diff
@@ -0,0 +1,6060 @@
+Arguments:
```
rwjblue: In general, this should not be committed to the repo

trentmwillis: Yep, good catch. I need to update my global ignore.

stefanpenner:
This babel plugin does not implement a cache key, causing the babel step to opt-out of its persistent-filter

PR #415, opened 2017-05-15
bump dependency-funnel

stefanpenner:


package.json
```diff
@@ -74,7 +74,7 @@
     "broccoli-babel-transpiler": "^5.6.1",
     "broccoli-concat": "^3.0.1",
     "broccoli-debug": "^0.6.0",
-    "broccoli-dependency-funnel": "^1.0.2",
+    "broccoli-dependency-funnel": "2.0.0",
```
rwjblue: @stefanpenner - The only breakage here was RE: dropping Node 0.12 support, right?

rwjblue: Also, can we use `^` instead of exact match?  Just in case we need to fix any bugs or something...

stefanpenner:
I dont understand this failure :(

stefanpenner:
NM, think I figured it out.... `files` list in broccoli-dependency-funnel wasn't up2date.

rwjblue:
Files trolls so hard 😩. I wish there was someway to prevent it from hitting so hard (perhaps an eslint plugin or something?)...

PR #421, opened 2017-05-25
Fixed transitionToExternal and replaceWithExternal to return thenable Transition objects

catrider:
Returning value from _externalAliasMethod, so that transitionToExternal and replaceWithExternal return a thenable Transition object

### DIFF
--- a/tests/acceptance/routeable-engine-demo-test.js
+++ b/tests/acceptance/routeable-engine-demo-test.js
@@ -277,6 +277,40 @@ test('transitionTo works properly within parent application', function(assert) {
   });
 });
 
+test('transitionToExternal transitions to the parent application from within an engine', function(assert) {
+  assert.expect(2);
+
+  visit('/routable-engine-demo/special-admin-blog-here/post/1');
+  click('.routable-post-transition-to-home-button');
+
+  andThen(() => {
+    assert.equal(currentURL(), '/');
+
+    visit('/routable-engine-demo/special-admin-blog-here/post/1');
+
+    andThen(() => {
+      assert.ok(this.application.$('.routable-post-transition-to-home-button').hasClass('transitioned-to-external'));
+    })
+  });
+});
+
+test('replaceWithExternal transitions to the parent application from within an engine', function(assert) {
+  assert.expect(2);
+
+  visit('/routable-engine-demo/special-admin-blog-here/post/1');
+  click('.routable-post-replace-with-home-button');
+
+  andThen(() => {
+    assert.equal(currentURL(), '/');
+
+    visit('/routable-engine-demo/special-admin-blog-here/post/1');
+
+    andThen(() => {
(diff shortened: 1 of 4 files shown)


tests/dummy/lib/ember-blog/addon/routes/post.js
```diff
@@ -13,6 +13,20 @@ export default Ember.Route.extend({
   actions: {
     goToChineseVersion() {
       this.transitionTo({ queryParams: { lang: 'Chinese' } });
+    },
+    transitionToHome() {
+      this.transitionToExternal('home')
+        .then(() => {
+          var postController = this.controllerFor(this.routeName);
+          postController.set('transitionedToExternal', true);
```
trentmwillis: I'd prefer we grab the controller for `home` and set a property on it so that we can avoid needing to `visit()` calls in the tests.

tests/acceptance/routeable-engine-demo-test.js
```diff
@@ -277,6 +277,40 @@ test('transitionTo works properly within parent application', function(assert) {
...
+
+  visit('/routable-engine-demo/special-admin-blog-here/post/1');
+  click('.routable-post-transition-to-home-button');
+
+  andThen(() => {
+    assert.equal(currentURL(), '/');
+
+    visit('/routable-engine-demo/special-admin-blog-here/post/1');
+
+    andThen(() => {
```
trentmwillis: Let's avoid nesting the test helpers as it makes the tests harder to read/follow.

tests/acceptance/routeable-engine-demo-test.js
```diff
@@ -277,6 +277,40 @@ test('transitionTo works properly within parent application', function(assert) {
   });
 });
 
+test('transitionToExternal transitions to the parent application from within an engine', function(assert) {
```
trentmwillis: We should also note that this test verifies the return value of `transitionToExternal`. Similarly for the `replaceWithExternal` test below.

tests/dummy/lib/ember-blog/addon/routes/post.js
catrider: Thanks for reviewing.  I don't believe we can get the controller for `home` in the engine since `home` lives in the parent?

trentmwillis: Ah, good point. Can we at least use normal navigation (e.g., `click`) to get back to the page? I'm not really a fan of multiple `visit` calls in a single test as that is not how an actual user will navigate the site.

catrider: 👍 . Done.

tests/acceptance/routeable-engine-demo-test.js
catrider: 👍 . Done.

tests/acceptance/routeable-engine-demo-test.js
catrider: 👍 . Done.

PR #422, opened 2017-05-25
Use cacheKeyForTree to improve the addon dedup

### LINKED ISSUE #416: Ember engines deduplication should use cacheKeyForTree instead of the addon name
Existing deduplication logic relies on the addon name to determine whether an addon should be deduplicated. This works OK for most addons, which returns the same trees irrespective of where in the addon hierarchy they are. However, some addons produce trees, which are specific to their parent or where in the dependency tree they are located. The deduplication mechanism for these addons may incorrectly dedup them, although the results are different between the parent and the child.

One such addon is [ember-cli-eyeglass](https://github.com/sass-eyeglass/ember-cli-eyeglass), which captures the assets produced by the eyeglass compiler and returns them later as part of the `treeForPublic` call. When the addon is dedup, the `treeForPublic` is never invoked, because it the parent contains the eyeglass compiler as a dependency as well.

The fix for this should be to change the existing logic for deduping addons based on cacheKeyForTree function, which should correctly identify if the tree is really duplicate. In case of `ember-cli-eyeglass` addon it will simply opt out of any dedup.

dnachev:
Attempt to fix #416 .

The change keeps the old logic for all invocations, which are not `treeFor`. For `treeFor`, the code will consult `cacheKeyForTree` to determine whether the trees from the host addon and the engine dependency addon are the same.

There is an open question of whether finding the corresponding host addon simply based on the name is enough. The host might have multiple instances of the same addon and we will only check for duplication against the first encountered one. Previously, the name was all cared about, so it worked just fine.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -296,12 +296,14 @@ module.exports = {
     };
 
     /**
-     * Gets a list of all the addons by name that are used by all hosts above
+     * Gets a map of all the addons that are used by all hosts above
      * the current host.
+     *
+     * The key is the name of the addon and the value is first encountered instance of this addon.
      */
-    options.ancestorHostAddonNames = function() {
-      if (this._hostAddonNames) {
-        return this._hostAddonNames;
+    options.ancestorHostAddons = function() {
+      if (this._hostAddons) {
+        return this._hostAddons;
       }
 
       var host = findHostsHost.call(this);
@@ -310,9 +312,9 @@ module.exports = {
         return [];
       }
 
-      var hostIsEngine = !!host.ancestorHostAddonNames;
+      var hostIsEngine = !!host.ancestorHostAddons;
 
-      var addonNames = hostIsEngine ? host.ancestorHostAddonNames().slice() : [];
+      var hostAddons = hostIsEngine ? Object.assign({}, host.ancestorHostAddons()) : {};
       var queue = hostIsEngine ? host.addons.slice() : host.project.addons.slice();
 
       // Do a breadth-first walk of the addons in the host, ignoring those that
@@ -321,32 +323,16 @@ module.exports = {
         var addon = queue.pop();
 
         if (addon.lazyLoading && addon.lazyLoading.enabled) { continue; }
-        if (addonNames.indexOf(addon.name) !== -1) { continue; }
 
-        addonNames.push(addon.name);
+        if (hostAddons[addon.name]) { continue; }
+
+        hostAddons[addon.name] = addon;
         queue.push.apply(queue, addon.addons);
       }
 
-      this._hostAddonNames = addonNames;
+      this._hostAddons = hostAddons;
 
-      return addonNames;
-    };
-
-    /**
-     * Gets a list of addons that are not included by an ancestor host based on
-     * the addon's name.
-     */
-    options.nonDuplicatedAddons = function() {
-      if (this._nonDuplicatedAddons) {
-        return this._nonDuplicatedAddons;
-      }
-
-      var hostAddonNames = this.ancestorHostAddonNames();
-      this._nonDuplicatedAddons = this.addons.filter(function(addon) {
-        return hostAddonNames.indexOf(addon.name) === -1;
-      });
-
-      return this._nonDuplicatedAddons;
+      return hostAddons;
     };
 
     /**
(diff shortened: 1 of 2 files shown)


lib/engine-addon.js
```diff
@@ -8,6 +8,7 @@ var writeFile = require('broccoli-file-creator');
 var Babel = require('broccoli-babel-transpiler');
 var concat = require('broccoli-concat');
 var DependencyFunnel = require('broccoli-dependency-funnel');
+var defaults = require('lodash.defaults');
```
trentmwillis: Can we just use `Object.assign` instead? I'm fairly certain we only need to support Node 4+ at this point due to our Ember-CLI support policy

node-tests/acceptance/build-test.js
```diff
@@ -398,5 +398,61 @@ describe('Acceptance', function() {
       output.doesNotContain(`engines-dist/${engineName}/assets/engine-vendor.js`, moduleMatcher(`${addonName}/templates/components/foo`));
       output.doesNotContain(`engines-dist/${engineName}/assets/engine-vendor.css`, cssCommentMatcher(`${addonName}.css`));
     }));
+
+    it('duplicate addons in lazy engines when tree is not the same', co.wrap(function* () {
```
trentmwillis: This would be clearer as: `duplicates an addon in lazy engines when the tree returned by the addon is different`

lib/engine-addon.js
```diff
@@ -358,12 +356,33 @@ module.exports = {
...
+            case 'treeFor':
+              var treeName = invokeArguments[0];
+              var hostAddonCacheKey = hostAddon.cacheKeyForTree(treeName);
+              var addonCacheKey = addon.cacheKeyForTree(treeName);
+              if (addonCacheKey != null && addonCacheKey === hostAddonCacheKey) {
+                // the addon specifies cache key and it is the same as host instance of the addon, skip the tree
+                return;
+              }
+              break;
+            case 'included':
```
trentmwillis: Why is `included` special cased here?

lib/engine-addon.js
```diff
@@ -281,6 +282,15 @@ var buildEngineStyleTree = memoize(function buildEngineStyleTree() {
   return this.debugTree(combinedEngineStylesAndDependencyStylesTree, 'engine-style:input');
 });
 
+/**
+ * Generates an identifier for an addon, which will be used when
+ * addons are deduped. Different instances of the same addon will
+ * be detected based on this identifier.
+ */
+function generateAddonId(addon) {
```
trentmwillis: Why do we need to call this `addonId` instead of just using `addon.name`? The function isn't doing anything special, and by introducing a new concept for identifying addons, it makes the code much harder to reason about.

lib/engine-addon.js
rwjblue: we can, but we'd have to bump to 0.6.0 (since 0.5.0 supported 0.12 when it released)

trentmwillis: But Ember-CLI 2.12 doesn't support Node 4, does it? I don't particularly care, I was just under the impression that our support policy was based on the Ember(-CLI) versions we supported. Regardless, now seems as good a time as any to drop 0.12.

rwjblue: Agreed all around. I had forgotten that we required Ember-cli@2.12 with ember-engines@0.5.0. Seems like 0.5 dropped node 0.12 support for us. 🎊

lib/engine-addon.js
dnachev: The initial idea was to encapsulate a better algorithm to determine the duplicate addon, which would be based on something more than just the name. However, considering that ember-cli doesn't have any notion of addon versioning in the same project, I think it was unnecessary. I will switch back to just use the name of the addon.

lib/engine-addon.js
dnachev: Just an explicit list of the cases, but I guess for only two cases makes for rather unordinary looking switch statement.

dnachev:
@trentmwillis I think I've addressed your comments.

I will try it out on our application to further proof it works once we have agreed this is the correct approach.

dnachev:
It seems our application still works with this fix. Is there anything else holding this out?

dnachev:
I would love another look at the changes to see whether there is anything more to be improved? @rwjblue @trentmwillis

PR #424, opened 2017-06-02
Update `lazyLoading` option to accept a hash

mike183:
Updates the `lazyLoading` option to accept a hash as its value. This is in preparation for a future PR to enable further options to be added to the lazy loading functionality.

Backwards compatibility has been retained by converting boolean values into a hash, e.g.

```js
lazyLoading: true

// Would be converted to

lazyLoading: {
  enabled: true
}
```

Not sure if this needs tests and if so, where they should actually be located as I had a quick look but couldn't find any existing tests for the flag.

cc/ @rwjblue

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -64,7 +64,7 @@ function findHost() {
   // Keep iterating upward until we don't have a grandparent.
   // Has to do this grandparent check because at some point we hit the project.
   do {
-    if (current.lazyLoading === true) { return current; }
+    if (current.lazyLoading.enabled === true) { return current; }
     app = current.app || app;
   } while (current.parent.parent && (current = current.parent));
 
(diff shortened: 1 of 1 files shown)


rwjblue:
restarted CI

lib/engine-addon.js
```diff
@@ -287,9 +287,9 @@ module.exports = {
 
     options.cacheKeyForTree = function(treeType) {
       // We do different things in the addon, public, and engine trees based on
-      // the value of `lazyLoading`, so we add it to the cache key
+      // the value of `lazyLoading.enabled`, so we add it to the cache key
       if (treeType === 'addon' || treeType === 'public' || treeType === 'engine') {
-        return calculateCacheKeyForTree(treeType, this, [ this.lazyLoading ]);
+        return calculateCacheKeyForTree(treeType, this, [ this.lazyLoading.enabled ]);
```
trentmwillis: I'm pretty sure we'd want to pass in the whole object here.

mike183: Fixed

mike183:
@trentmwillis Thanks for the review, I've addressed your issue.

Also can't figure out why CI would be failing?

PR #425, opened 2017-06-03
Introduce `lazyLoading.includeRoutesInApplication` option

mike183:
Introduces a new `lazyLoading.includeRoutesInApplication` option which enables lazy engines to output their `routes.js` file to a stand-alone file rather than bundling it into the hosts `vendor.js` file.

This will enable engines to be discovered dynamically at runtime rather than requiring engines to be declared statically at build time.

Still needs some tests which I will do once others have reviewed.

**Usage:**

```js
// Prevent routes being bundled into hosts vendor.js
lazyLoading: {
  enabled: true,
  includeRoutesInApplication: false
}
```

**TODO:**
- [x] Tests

cc/ @rwjblue @trentmwillis

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -185,6 +185,29 @@ var buildEngineJSTreeWithoutRoutes = memoize(function buildEngineJSTreeWithoutRo
   });
 });
 
+var buildEngineRoutesJSTree = memoize(function buildEngineRoutesJSTree() {
+  // Get complete engine JS tree
+  var engineAppTree = buildEngineJSTree.call(this);
+  // Separate routes
+  var engineRoutesTree = new DependencyFunnel(engineAppTree, {
+    include: true,
+    entry: this.name+'/routes.js',
+    external: ['ember-engines/routes']
+  });
+  // If babel options aren't defined, we need to transpile the modules.
+  if (!this.options || !this.options.babel) {
+    engineRoutesTree = processBabel(engineRoutesTree);
+  }
+  // Concatenate routes.js and its dependencies into a single file.
+  engineRoutesTree = concat(engineRoutesTree, {
+    allowNone: true,
+    inputFiles: ['**/*.js'],
+    outputFile: 'engines-dist/' + this.name + '/assets/routes.js'
+  });
+  // Return concatenated JS tree
+  return engineRoutesTree;
+});
+
 var buildVendorJSWithImports = memoize(function buildVendorJSWithImports(concatTranspiledVendorJSTree) {
   var externalTree = buildExternalTree.call(this);
   var combined = mergeTrees([externalTree, concatTranspiledVendorJSTree].filter(Boolean), { overwrite: true });
@@ -487,10 +510,15 @@ module.exports = {
         this.nonDuplicatedAddonInvoke('included', [this]);
       };
 
+      // The treeForEngine method constructs and returns a tree that represents
+      // the engines routes.js file and its dependencies. This is used later to
+      // promote the engines routes.js to the host.
+      //
+      // If the lazyLoading.includeRoutesInApplication option is false, we don't
+      // want to promote the routes into the host.
       this.treeForEngine = function() {
-        // Unless includeRoutesInApplication is false, if this engine or any of
-        // its parents are lazy we need to promote its routes.
-        if (!this._hasLazyAncestor || this.lazyLoading.includeRoutesInApplication === false) { return; }
+        var extractRoutes = this._hasLazyAncestor && this.lazyLoading.includeRoutesInApplication === false;
+        if (extractRoutes) { return; }
 
         // The only thing that we want to promote from a lazy engine is the
         // routes.js file and all of its dependencies, which is why we build the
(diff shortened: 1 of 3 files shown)


lib/engine-addon.js
```diff
@@ -488,8 +488,9 @@ module.exports = {
       };
 
       this.treeForEngine = function() {
-        // If this engine is lazy or any of its parents are lazy we need to promote its routes.
-        if (!this._hasLazyAncestor) { return; }
+        // Unless includeRoutesInApplication is false, if this engine or any of
+        // its parents are lazy we need to promote its routes.
+        if (!this._hasLazyAncestor || this.lazyLoading.includeRoutesInApplication === false) { return; }
```
trentmwillis: This line and the comment above it are hard to grok. I think this is likely because there isn't clear indication upfront about what this function does (and not your fault).

My suggestion: we should add a block comment to the function stating that `treeForEngine` returns a tree representing `routes.js` and its dependency graph.

Then, we can make this block more understandable by using a descriptive variable name:

```js
var extractRoutes = this._hasLazyAncestor && this.lazyLoad.includeRoutesInApplication !== false;
if (!extractRoutes) { return; }
```

lib/engine-addon.js
```diff
@@ -664,10 +665,24 @@ module.exports = {
 
         var engineJSTree = buildEngineJSTreeWithoutRoutes.call(this);
 
+        // If includeRoutesInApplication is false, separate the routes.js file
+        // and its dependencies
+        if (this.lazyLoading.includeRoutesInApplication === false) {
+          var engineAppTree = buildEngineJSTree.call(this);
+          var engineRoutesTree = new DependencyFunnel(engineAppTree, {
```
trentmwillis: This variable should be declared outside this block.

lib/engine-addon.js
```diff
@@ -664,10 +665,24 @@ module.exports = {
 
         var engineJSTree = buildEngineJSTreeWithoutRoutes.call(this);
 
+        // If includeRoutesInApplication is false, separate the routes.js file
+        // and its dependencies
+        if (this.lazyLoading.includeRoutesInApplication === false) {
```
trentmwillis: Can we create a variable `var separateRoutes = this.lazyLoading.includeRoutesInApplication === false;`?

I think it would make the code easier to read and allows us to only perform this check in one location.

lib/engine-addon.js
```diff
@@ -696,17 +720,20 @@ module.exports = {
...
-        );
+        var finalMergeTrees = [
+          publicRelocated,
+          addonsEnginesPublicTreesMerged,
+          otherAssets,
+          finalStylesTree,
+          vendorJSImportTree,
+          concatEngineTree
+        ];
+        if (this.lazyLoading.includeRoutesInApplication === false) {
```
trentmwillis: Do you think we could combine all of the different `if` blocks added into one block? In addition, I think we could extract this all out into a separate function that returns a tree to merge in.

lib/engine-addon.js
mike183: Added a block comment and a descriptive variable name to try and make this function a bit easier to understand.

lib/engine-addon.js
mike183: Done

lib/engine-addon.js
mike183: No longer necessary due to the refactor.

lib/engine-addon.js
mike183: Followed your suggestion of refactoring this into a single function that returns the tree to be merged in.

mike183:
Thanks for the review @trentmwillis, next up is some tests, then this should be good to go.

mike183:
Just pushed a test for this, I feel like the test could be simplified a bit but it serves the purpose for now.

mike183:
Is there anything else needed to get this merged?

lib/engine-addon.js
```diff
@@ -185,6 +185,29 @@ var buildEngineJSTreeWithoutRoutes = memoize(function buildEngineJSTreeWithoutRo
...
+    engineRoutesTree = processBabel(engineRoutesTree);
+  }
+  // Concatenate routes.js and its dependencies into a single file.
+  engineRoutesTree = concat(engineRoutesTree, {
+    allowNone: true,
+    inputFiles: ['**/*.js'],
+    outputFile: 'engines-dist/' + this.name + '/assets/routes.js'
+  });
+  // Return concatenated JS tree
+  return engineRoutesTree;
```
rwjblue: Can you wrap this in a `this.debugTree` so that we can have a tad easier time at debugging when things go wrong?

Specifically, I'm suggesting:

```js
return this.debugTree(engineRoutesTree, 'routes:output');
```

mike183: Done.

CI errors looks unrelated and may just need a restart.

lib/engine-addon.js
```diff
@@ -487,9 +510,15 @@ module.exports = {
...
+      // The treeForEngine method constructs and returns a tree that represents
+      // the engines routes.js file and its dependencies. This is used later to
+      // promote the engines routes.js to the host.
+      //
+      // If the lazyLoading.includeRoutesInApplication option is false, we don't
+      // want to promote the routes into the host.
       this.treeForEngine = function() {
-        // If this engine is lazy or any of its parents are lazy we need to promote its routes.
-        if (!this._hasLazyAncestor) { return; }
+        var extractRoutes = this._hasLazyAncestor && this.lazyLoading.includeRoutesInApplication === false;
```
mike183: I think you're right, updated.

rwjblue: Good catch!

PR #427, opened 2017-06-06
blueprint updates to conform with latest cli style

kellyselden:


blueprints/in-repo-engine/routable-files/lib/__name__/addon/templates/application.hbs
```diff
@@ -1 +1 @@
-{{outlet}}
+{{outlet}}
```
trentmwillis: What's the reasoning for removing the new line? I just ask since most projects I work on have a style rule to have the new line.

kellyselden: The default ember-cli blueprint prefers no end newlines in templates
https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.editorconfig#L17

trentmwillis: Meh, not a fan of the inconsistency, but if that's the default for apps, should be the same here.

Issue #376, opened 2017-03-21
`treeForAddon` not calling super in 0.5.0


ming-codes:
Specifically, this line https://github.com/ember-engines/ember-engines/blob/v0.4.0/lib/engine-addon.js#L297 is gone in 0.5.0

[diff here](https://github.com/ember-engines/ember-engines/compare/v0.4.0...v0.5.0#diff-1c2c32318ef6ba8d9be4e7b085f8ecb5L297)

ming-codes:
A few other hooks seems to also not calling the child class's `treeFor` hooks. Is this intentional?

rwwagner90:
@ming-codes I just updated to 0.5.0 and ember-cli-sass broke. I assume this is the culprit.

rwjblue:
@rwwagner90 - I doubt it.  Can you open a separate issue and possibly put together a reproduction?

trentmwillis:
To make sure we're on the same page, the issue is that a custom `treeForAddon` hook in your Engine is not being called, correct? The `treeForAddon` hook within nested addons should still be getting triggered properly based on the info provided.

rwwagner90:
@rwjblue unsure if the issue is with engines or if ember-cli-sass needs to be updated to conform to the new way of doing things. I haven't actually done much with engines myself, just trying to update all our apps/addons/engines to 2.12, and when our engine broke on 2.12 with an error something to the effect of not knowing what `define` was and saying `export` was unexpected or something, I figured updating engines would fix the issue, which I think it did, but then ember-cli-sass broke. You can see the hooks used there here https://github.com/aexmachina/ember-cli-sass/blob/master/index.js

rwwagner90:
@rwjblue @trentmwillis new issue created, in case these are not related to the same issue https://github.com/ember-engines/ember-engines/issues/378

ming-codes:
> To make sure we're on the same page, the issue is that a custom treeForAddon hook in your Engine is not being called, correct?

Yup.

```
EmberAddon.extend({
  treeForAddon() { } // <!-- this is not called
});
```

toddjordan:
I'd like to add that my company has a fairly large scale ember engines deployment and this issue is currently keeping us from moving to ember ^2.12.  A number of our engines leverage `treeForAddon` to do less compilation (we keep less files in a pod structure for components), so we are seeing after upgrade that our engine screens are unstyled.  

[...]

PR #428, opened 2017-06-06
[WIP] call original treeForAddon (again)

toddjordan:
Initial stab at #376 
Plan to add some tests around the changed behavior.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -502,11 +502,13 @@ module.exports = {
 
       // Replace `treeForAddon` so that we control how this engine gets built.
       // We may or may not want it to be combined like a default addon.
+      var originalTreeForAddon = this.treeForAddon;
       this.treeForAddon = function() {
         if (this.lazyLoading === true) { return; }
 
         // NOT LAZY LOADING!
         // This is the scenario where we want to act like an addon.
+        var originalEngineAddonTree = originalTreeForAddon.apply(this, arguments);
         var engineCSSTree = buildEngineStyleTree.call(this);
 
         if (useDeprecatedIncorrectCSSProcessing) {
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm.  What are y'all doing custom in `treeForAddon`? I'm wondering if the functionality added in https://github.com/ember-engines/ember-engines/pull/357 would meet your needs...

toddjordan:
@rwjblue Possibly.  What we are primarily doing in some of our engines is compiling less files that exist in component pods in `/addon/components/*`, so if we had access to the addonTree in another hook, such as treeForStyles, then we could adapt that code.

toddjordan:
@rwjblue How would I get the addon tree in this case?  it looks like `this.options.trees.addon' just gives me the String `addon`.

toddjordan:
Was able to put the following code into one of our engines (without this fix) and have it work:

```javascript
  treeForVendor(tree) {
    let addonTree = this.treeGenerator(path.resolve(this.root, this.options.trees.addon));
    addonTree = debug(addonTree, {name: 'addonTree'});
    let compiledLessTree = new LessCompiler(addonTree, 'styles/addon.less', `${this.name}.css`);
    compiledLessTree = debug(compiledLessTree, {name: 'compiledLess'});
    return new TreeMerger([tree, compiledLessTree]);
  }
```

Using treeGenerator feels a bit hacky though.  Is this what you were thinking about?

toddjordan:
Since this PR didn't seem to get much traction, we went with the `treeForVendor` approach shown above instead.  The thing I'm concerned with is that `treeGenerator` is private in the api doc.  https://ember-cli.com/api/classes/Addon.html#method_treeGenerator
Should we elevate this to public or is there another/better way?

cc @rwjblue

nmcclay:
What is the status of this PR? Its been hanging out for a year now and I'm wondering when/if we can use the vanilla addon treeForAddon hook?

stefanpenner:
@nmcclay will add this to my long queue of making engines ready for 1.0

nmcclay:
Nice, thanks so much @stefanpenner !

PR #434, opened 2017-06-13
Make addons accept asset objects /w environment properties

buschtoens:
Based on https://github.com/ember-engines/ember-engines/pull/252

@trentmwillis Is this what you meant by "return early"?


trentmwillis:
Yep, thank you!

rwjblue:
released as v0.5.7

PR #443, opened 2017-07-07
Assert if link-to is used within a routeless engine

### LINKED ISSUE #272: Route-less Engine link-to scoping
Currently, within a routable Engine `{{link-to}}` gets "scoped" to the Engine's `mountPoint`. However, this same behavior is not present for route-less Engines since they do not currently have a `mountPoint` since it is defined by the Router's DSL.

The question thus is: _should route-less Engines have `{{link-to}}` scoped in the same way as routable Engines?_

I am inclined to say yes, primarily for the two following reasons:

1. It makes behavior consistent across types of Engines.
2. It would force route-less Engines to use `link-to-external` which means explicit contracts about linking could be made between the host and the Engine.
3. Keeps route-less Engines from coupling to their host's route structure.

Note that this would mean that you effectively can't use `{{link-to}}` in route-less Engines, since they have no child routes.

Any other thoughts?

mike183:
Adds an assertion if `{{link-to}}` is used within a routeless engine due to it not being supported.

Not sure if we still want to scope `{{link-to}}`'s within a routeless engine, or if the assertion alone is enough.

Closes #272

### DIFF
--- a/addon/components/link-to-component.js
+++ b/addon/components/link-to-component.js
@@ -17,7 +17,7 @@ export default LinkComponent.extend({
     let owner = getOwner(this);
 
     assert(
-      `You attempted to use a link-to within a routeless engine, this is not supported.`,
+      `You attempted to use {{link-to}} within a routeless engine, this is not supported. Use {{link-to-external}} to construct links within a routeless engine. See http://ember-engines.com/guide/linking-and-external-links for more info.`,
       owner.mountPoint !== undefined
     );
 
--- a/addon/components/link-to-component.js
+++ b/addon/components/link-to-component.js
@@ -6,7 +6,8 @@ const {
   getOwner,
   get,
   set,
-  typeOf
+  typeOf,
+  assert
 } = Ember;
 
 export default LinkComponent.extend({
@@ -14,6 +15,11 @@ export default LinkComponent.extend({
     this._super(...arguments);
 
     let owner = getOwner(this);
+
+    assert(
+      `You attempted to use a link-to within a routeless engine, this is not supported.`,
+      owner.mountPoint !== undefined
+    );
 
     if (owner.mountPoint) {
       // Prepend engine mount point to targetRouteName


addon/components/link-to-component.js
```diff
@@ -15,6 +16,11 @@ export default LinkComponent.extend({
 
     let owner = getOwner(this);
 
+    assert(
+      `You attempted to use a link-to within a routeless engine, this is not supported.`,
```
trentmwillis: Do we have access to the Engine's `modulePrefix` so that we can let the user know which Engine this occurred in (ideally it'll be caught by the developer, but ya never know)? Also, we should probably mention something like:

`In order to construct links in a routeless, use {{link-to-external}}. See http://ember-engines.com/guide/linking-and-external-links for more info.`

mike183: I had considered this briefly but wasn't sure how to gain access to the modulePrefix.

@rwjblue Any ideas?

trentmwillis: Does the `owner` have easy access to it? If it's not straightforward, we can worry about it if it becomes an issue.

mike183: Ok, I've not tried it yet as I'm on mobile, but we may be able to lookup the config using the code below, then access the `modulePrefix` from there. 

```js
const config = owner.lookup('config:environment');
```

If we can use this to access the module prefix, I think it could be a good idea to scope the link-to's as well as add the assertion. This way we're effectively rendering the link-to component useless with the scope of a routeless engine.

Would you agree?

trentmwillis: :+1: Seems reasonable to me, though the assertion should make it useless on its own (as it causes an exception).

mike183: So it turns out that the above lookup didn't work as  `config:environment` isn't listed in the owners registry. I'm sure it will be possible to retrieve the `modulePrefix`, though I'm unsure how exactly.

Either way, I've updated the assertions message to include a note about the `{{link-to-external}}` component and a link to its documentation.

mike183:
Is there anything else required to land this?

trentmwillis:
Do you think you could add a test case for this? I was originally waiting for Rob to review, but I think if we have a test we should be good to go.

mike183:
I'll look at adding a test when I get a minute, though it may be next week now.

rwjblue:
sorry, been on vacation (will be back in the saddle next week)

PR #446, opened 2017-07-12
fix: make sure externalRoutes are only handled if dependencies exist

knownasilya:


knownasilya:
Looks like there is already a test https://github.com/ember-engines/ember-engines/blob/master/tests/unit/engine-instance-test.js#L45 maybe it doesn't do it the right way?

trentmwillis:
@knownasilya I think it is missing one of the potential branches. We need to test for the [`engineConfiguration` being defined](https://github.com/knownasilya/ember-engines/blob/191988d4a29d39ba1a2500aa46d345d37f55b226/addon/-private/engine-instance-ext.js#L60-L62) but the [`engineDependencies` not being defined](https://github.com/knownasilya/ember-engines/blob/191988d4a29d39ba1a2500aa46d345d37f55b226/addon/-private/engine-instance-ext.js#L63-L65).

rwjblue:
@knownasilya - Had a chance to add the test yet?

knownasilya:
Not yet, I'll try to get to it this weekend. Thanks for the reminder.

rwjblue:
@knownasilya - Mind adding a test?

knownasilya:
So the tests are actually there. I'm not sure why they didn't catch this issue. The first test doesn't define dependencies in the parent app, and that test passes with this change. There is a second test, and that one does define dependencies: https://github.com/ember-engines/ember-engines/blob/master/tests/unit/engine-instance-test.js#L75

knownasilya:
I added a test that resets `app.engines` to `undefined`.

rwjblue:
Thanks for working through this! Can you double check and confirm the new test failed prior to the change? (I mostly want to doubly ensure this is testing the right scenario)

trentmwillis:
That was sneaky, thanks for figuring all this out @knownasilya!

PR #452, opened 2017-07-18
add test for component from lazy engine

kellyselden:
I'm trying to recreate an issue I'm having in my project, but so far I can't repro in your suite.

This seems like a good starting point.


kellyselden:
Nothing. good call.

PR #455, opened 2017-07-20
compiledEngineCSSTree should not be falsy

lynchbomb:
`this.compiledStyles` can return a falsy value as such `this.debugTree` should not be called in this instance.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -564,7 +564,10 @@ module.exports = {
           engineCSSTree = this._treeFor('addon-styles');
         }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Ya, I think we should fix this upstream in broccoli-debug.

rwjblue:
https://github.com/broccolijs/broccoli-debug/pull/9 should fix this, can you review?

rwjblue:
` broccoli-debug 0.6.3 published 🎉`

I think updating to broccoli-debug@0.6.3 should be all that is needed now.

trentmwillis:
@rwjblue awesome, I took a look at the broccoli-debug update :+1: exactly what I was thinking

PR #457, opened 2017-07-27
[BUGFIX release]: ancestorHostAddons return value

alexander-alvarez:
Without this, you cannot build engines with who's names are built-in array methods such as "filter".
The return type of the function is an `Object` not an `Array`


knownasilya:
Ping @rwjblue @trentmwillis

PR #464, opened 2017-08-07
Update to latest ember-cli@2.14.1 blueprint.

rwjblue:
Note: this includes updating to Node 4+, which will require a 0.6.0.


package.json
```diff
@@ -89,8 +91,10 @@
     "eslint": "^3.19.0",
     "exists-sync": "0.0.3",
     "lodash": "^4.12.0",
-    "lodash.defaultsdeep": "^4.5.1",
-    "mocha-eslint": "^3.0.1"
```
rwjblue: `mocha-eslint` was mistakenly in `dependencies`, this just moves it `devDependencies`.

rwjblue:
Tracking the caching failure that required 8ffa49f in https://github.com/ember-cli/ember-cli-htmlbars-inline-precompile/issues/95.

rwjblue:
Turns out that we were already using ember-cli-babel@6 (since version 0.5.3) so this won't need to be a breaking change release at all...

PR #468, opened 2017-08-16
Updated contentFor method to support addons

robabby:


robabby:
Disregard this PR.  I made it on accident.

Issue #435, opened 2017-06-13
"hostAddon.cacheKeyForTree is not a function 


bbtb1982:
We're getting this error when building our project with v5.5 - v5.8 version 5.4 builds fine.

hostAddon.cacheKeyForTree is not a function
TypeError: hostAddon.cacheKeyForTree is not a function
    at .../node_modules/ember-engines/lib/engine-addon.js:382:49
    at Array.map (native)
    at Class.options.nonDuplicatedAddonInvoke (.../ember-deployment/node_modules/ember-engines/lib/engine-addon.js:372:26)
    at Class.buildVendorTree (.../node_modules/ember-engines/lib/engine-addon.js:129:36)
    at Class._memoize (.../node_modules/ember-engines/lib/utils/memoize.js:10:38)
    at Class.buildCompleteJSTree (.../node_modules/ember-engines/lib/engine-addon.js:255:36)
    at Class._memoize (.../node_modules/ember-engines/lib/utils/memoize.js:10:38)
    at Class.treeForEngine (.../node_modules/ember-engines/lib/engine-addon.js:540:50)
    at Class.treeFor (.../node_modules/ember-engines/lib/engine-addon.js:928:25)

here is our NPM dependencies

  "devDependencies": {
    "broccoli-asset-rev": "^2.4.5",
    "ember-ajax": "^2.4.1",
    "ember-cli": "2.12.1",
    "ember-cli-app-version": "^2.0.0",
    "ember-cli-babel": "^5.1.7",
    "ember-cli-cdn": "2.1.0",
    "ember-cli-content-security-policy": "0.6.1",
    "ember-cli-dependency-checker": "^1.3.0",
    "ember-cli-deploy": "^1.0.0",
    "ember-cli-deploy-build": "^1.0.0",
    "ember-cli-deploy-display-revisions": "^1.0.0",
    "ember-cli-deploy-gzip": "^1.0.0",
    "ember-cli-deploy-revision-data": "^1.0.0",
    "ember-cli-deploy-s3": "^1.0.0",
    "ember-cli-deploy-s3-index": "^1.0.0",
    "ember-cli-deploy-slack": "^1.0.0",
    "ember-cli-document-title": "0.3.3",
    "ember-cli-eslint": "^3.0.0",
    "ember-cli-g-maps": "0.5.10",
    "ember-cli-htmlbars": "^1.1.1",
    "ember-cli-htmlbars-inline-precompile": "^0.3.6",
[...]

PR #469, opened 2017-08-21
Fix: Default to no caching if cacheKeyForTree is not available in host

akashdsouza:
As per https://github.com/ember-engines/ember-engines/issues/435#issuecomment-323118100


rwjblue:
Thank you!

rwjblue:
Published as v0.5.12

Issue #474, opened 2017-08-31
engine boot promise is not chained to the promise of the engine


krisselden:
https://github.com/ember-engines/ember-engines/blob/7f8b02fee1b0b40200c5411c59f692759d5f660e/addon/-private/router-ext.js#L254

This leads to errors being swallowed.  If the engine fails to boot, it fails downstream of this error when trying to transition into the engines route but without the original cause.

PR #477, opened 2017-09-01
Ensure the engine instances promise properly rejects if boot fails.

rwjblue:
Fixes https://github.com/ember-engines/ember-engines/issues/474.


rwjblue:
See description in #478 for why the testing for this has proven overly difficult.  Some changes in Ember and/or router.js upstream will be required to make this testable.  Did smoke testing locally to confirm that this properly entangles the rejection with the failed boot promise.

PR #481, opened 2017-09-05
Retry engine on assets error

### LINKED ISSUE #471: Retry engine on assets error
Hi.
I don't know if this is really a bug or is the correct implementation but I think I found a cache problem when dealing with the lazy engines.

### How to reproduce the error:
1. Visit an application with a lazy load engine (don't load it yet).
2. In developer tools **enable** the **Offline** mode in the Network panel.
3. Navigate to the lazy engine, the assets _engine.js_ and _engine-vendor.js_ must fail at this point.
4. **Disable** the **Offline** mode.
5. Try to access again to the lazy engine, at this point the engine is inaccesible because the previous errored promise is cached [router-ext.js#L206](https://github.com/ember-engines/ember-engines/blob/master/addon/-private/router-ext.js#L206).

### ¿Why I think the current cache is a bug?
When developing a mobile app the user sometimes can have troubles with its connection, maybe because the wifi is lost or there is not enought network coverage.

To fix the problem I've created a branch named *error-cache* in both [ember-engines](https://github.com/BBVAEngineering/ember-engines/tree/error-cache) and [ember-asset-loader](https://github.com/BBVAEngineering/ember-asset-loader/tree/error-cache) repos (only 2 or 3 commits).
The fix needs changes in 2 repos and I'm not sure if this changes can have side effects, thats why I opened an issue instead a pull-request.

¿What do you think?

josex2r:
This pull tries to fix #471.

There are some changes in the acceptance tests that I don't like but I didn't find any other way to implement them.
I had to remove the exception adapter (`Ember.Test.adapter.exception`) because the router creates a new promise without an error handler ([L47:router.js](https://github.com/tildeio/router.js/blob/816c11f99bb169300ec98dbdb613bc2c0cf6c175/lib/router/handler-info.js#L47)). When I force an error in a transition, the exception is caught by the `ember-test` adapter making the test fail.
¿Do you know prevent the test failing?

Also I changed the `package.json` to trigger the build in Travis with the changes that I've done in [ember-asset-loader](https://github.com/ember-engines/ember-asset-loader/pull/57). This change should be removed.

### DIFF
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -215,7 +215,7 @@ Router.reopen({
       enginePromise = this._assetLoader.loadBundle(name).then(
         () => this._registerEngine(name),
         (error) => {
-          delete enginePromises[name][instanceId];
+          enginePromises[name][instanceId] = undefined;
           throw error;
         }
       );
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -216,7 +216,7 @@ Router.reopen({
         () => this._registerEngine(name),
         (error) => {
           delete enginePromises[name][instanceId];
-          return RSVP.reject(error);
+          throw error;
         }
       );
     }
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -212,7 +212,13 @@ Router.reopen({
       enginePromise = RSVP.resolve();
     } else {
       // The Engine is not loaded and has no Promise
-      enginePromise = this._assetLoader.loadBundle(name).then(() => this._registerEngine(name));
+      enginePromise = this._assetLoader.loadBundle(name).then(
+        () => this._registerEngine(name),
+        (error) => {
+          delete enginePromises[name][instanceId];
+          return RSVP.reject(error);
+        }
+      );
     }
 
     return enginePromises[name][instanceId] = enginePromise.then(() => {
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "broccoli-merge-trees": "^1.0.0",
     "broccoli-test-helper": "^1.1.0",
     "calculate-cache-key-for-tree": "^1.0.0",
-    "ember-asset-loader": "git+https://github.com/BBVAEngineering/ember-asset-loader.git#prevent-caching-errors",
+    "ember-asset-loader": "^0.3.1",
     "ember-cli-babel": "^6.7.1",
     "ember-cli-preprocess-registry": "^3.0.0",
     "ember-cli-string-utils": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "broccoli-merge-trees": "^1.0.0",
     "broccoli-test-helper": "^1.1.0",
     "calculate-cache-key-for-tree": "^1.0.0",
-    "ember-asset-loader": "^0.3.0",
+    "ember-asset-loader": "git+https://github.com/BBVAEngineering/ember-asset-loader.git#prevent-caching-errors",
     "ember-cli-babel": "^6.7.1",
     "ember-cli-preprocess-registry": "^3.0.0",
     "ember-cli-string-utils": "^1.0.0",
--- a/tests/acceptance/lazy-routeable-engine-test.js
+++ b/tests/acceptance/lazy-routeable-engine-test.js
@@ -1,3 +1,4 @@
+/* global define, window */
 import Ember from 'ember';
 import { test } from 'qunit';
 import moduleForAcceptance from '../../tests/helpers/module-for-acceptance';
@@ -5,6 +6,7 @@ import App from '../../app';
 
 const { RSVP } = Ember;
 const SEPARATORS = /\/|\\/;
+const nope = () => {};
 
 moduleForAcceptance('Acceptance | lazy routable engine', {
   beforeEach() {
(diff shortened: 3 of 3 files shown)


addon/-private/router-ext.js
```diff
@@ -212,7 +212,13 @@ Router.reopen({
       enginePromise = RSVP.resolve();
     } else {
       // The Engine is not loaded and has no Promise
-      enginePromise = this._assetLoader.loadBundle(name).then(() => this._registerEngine(name));
+      enginePromise = this._assetLoader.loadBundle(name).then(
+        () => this._registerEngine(name),
+        (error) => {
+          delete enginePromises[name][instanceId];
```
rwjblue: Since we only do truthy checks we can likely avoid using `delete` here:

```js
enginePromises[name][instanceId] = undefined;
```

package.json
```diff
@@ -80,7 +80,7 @@
     "broccoli-merge-trees": "^1.0.0",
     "broccoli-test-helper": "^1.1.0",
     "calculate-cache-key-for-tree": "^1.0.0",
-    "ember-asset-loader": "^0.3.0",
+    "ember-asset-loader": "git+https://github.com/BBVAEngineering/ember-asset-loader.git#prevent-caching-errors",
```
rwjblue: Just leaving a note here so we don't forget to update this to a released version once the other PR is merged

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -14,28 +16,23 @@ moduleForAcceptance('Acceptance | lazy routable engine', {
 
     // We stub out the loader methods so that we can verify what they're doing.
     const module = this;
+    this.loader = this.application.__container__.lookup('service:asset-loader');
```
rwjblue: In general it is better to use `this.application.__deprecatedInstance__` instead (as represents an object that is already public API).

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -66,6 +63,40 @@ test('it should pause to load JS and CSS assets on deep link into a lazy Engine'
...
+
+  define('dummy/routes/application', () => (Ember.Route.extend({
+      actions: {
+        error: nope
+      }
+    })
+  ));
+
+  const adapterException = Ember.Test.adapter.exception;
+  const loader = this.application.__container__.lookup('service:asset-loader');
```
rwjblue: I think you can use `this.loader` here (since you saved it off on the test context above in the `beforeEach`).

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -66,6 +63,40 @@ test('it should pause to load JS and CSS assets on deep link into a lazy Engine'
...
+  const jsLoader = loader.__assetLoaders.js;
+  const cssLoader = loader.__assetLoaders.css;
+  const failLoader = () => RSVP.reject('rejected');
+
+  Ember.Test.adapter.exception = nope;
```
rwjblue: This is unfortunate, but I believe it is acceptable.  I am not sure what a valid rejection handler would actually look like in router.js (from its perspective there is no real recovery AFAICT) so doing this to avoid failing the test while we confirm the underlying behavior seems fine.

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -74,6 +105,40 @@ test('it should pause to load JS and CSS assets on an initial transition into a
   andThen(() => verifyInitialBlogRoute(assert, this.loadEvents, this.application));
 });
 
+test('it should pause to load JS and CSS assets on deep link into a lazy Engine after retrying the transition', function(assert) {
```
rwjblue: Its not immediately obvious to me how this test is supposed to differ from the other test being added just above, can you modify the test descriptions to make it clearer?

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -66,6 +63,40 @@ test('it should pause to load JS and CSS assets on deep link into a lazy Engine'
...
+  visit('/routable-engine-demo/blog/new');
+
+  andThen(() => {
+    loader.defineLoader('js', jsLoader);
+    loader.defineLoader('css', cssLoader);
+    visit('/routable-engine-demo/blog/new');
+
+    andThen(() => {
+      verifyInitialBlogRoute(assert, this.loadEvents, this.application);
+      Ember.Test.adapter.exception = adapterException;
```
rwjblue: This cleanup should be in an `afterEach`, that guarantees that it is properly reset if any of these tests fail...

tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -66,6 +63,40 @@ test('it should pause to load JS and CSS assets on deep link into a lazy Engine'
   andThen(() => verifyInitialBlogRoute(assert, this.loadEvents, this.application));
 });
 
+test('it should pause to load JS and CSS assets on deep link into a lazy Engine after retrying the transition', function(assert) {
+  assert.expect(7);
+
+  define('dummy/routes/application', () => (Ember.Route.extend({
+      actions: {
+        error: nope
```
rwjblue: In general, its easier for me to grok the test if this was just `error() { }` (which happens to be ever so slightly less characters anyways 😝 )...

tests/acceptance/lazy-routeable-engine-test.js
rwjblue: Can you swap this to `() => {}` directly though?  Its just a little easier to grok to me...

tests/acceptance/lazy-routeable-engine-test.js
josex2r: Nice! didn't know that

tests/acceptance/lazy-routeable-engine-test.js
josex2r: Ok!

tests/acceptance/lazy-routeable-engine-test.js
josex2r: Something like `it should pause to load JS and CSS assets if previous deep link into a lazy Engine has failed` could be better?

rwjblue: Yes, that sounds better to me

package.json
trentmwillis: This can be updated now.

josex2r:
Yes, to retry the load its necessary to repeat the transition.
Here is an example of how it works with `transitionTo`:

transition-to

rwjblue:
Confirm.  Automated retry is still something that someone could do (e.g. from their own `error` action or something), but IMHO is not something that we want to come out of the box.

rwjblue:
Released as v0.5.13

trentmwillis:
Awesome, thanks for working on this @josex2r!

PR #492, opened 2017-10-07
Bump asset loader

mfeckie:
Simple bump of asset loader to include bugfix to avoid full page refresh.


rwjblue:
Hmm, seems like CI isn't setting up our test only in-repo addons correctly.  Will need to dig in to figure out why...

rwjblue:
Merged in master to get the fixes from @trentmwillis in #493

PR #493, opened 2017-10-07
Fix CI

trentmwillis:
Hopefully this fixes the issues with CI. As far as I can tell, Yarn is no longer running the `prepublish` script on install which is responsible for linking the test engines to their proper locations.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -1,6 +1,7 @@
 'use strict';
 
 module.exports = {
+  useYarn: true,
   scenarios: [
     {
       name: 'ember-lts-2.12',
--- a/node-tests/lint-test.js
+++ b/node-tests/lint-test.js
@@ -11,6 +11,7 @@ lint([
   'config',
   'lib',
   'node-tests',
-  'tests',
+  'tests/**/*.js',
+  '!tests/**/node_modules/**/*.js',
   'vendor',
 ]);
--- a/yarn.lock
+++ b/yarn.lock
@@ -2106,9 +2106,9 @@ ember-ajax@^3.0.0:
   dependencies:
     ember-cli-babel "^6.0.0"
 
-ember-asset-loader@^0.3.0:
-  version "0.3.0"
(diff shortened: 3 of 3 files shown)


trentmwillis:
Apparently I was wrong. Going to try removing ember-cli-dependency-checker and see if the tests pass, as that is preventing them from running at all.

rwjblue:
Thanks for working on this @trentmwillis.

trentmwillis:
So it runs without the dep checker, but the tests fail for not having the right output files, which would be a byproduct of what the dep checker was warning against. Trying some other tweaks now...

rwjblue:
Does it have the same issues locally?  I'm wondering if we are being trolled by recent changes to ember-try (whcih was to not use `yarn` without a `useYarn: true` flag) or possibly by getting npm@5 incompatibilities...

trentmwillis:
I have now fixed the Ember tests, which was due to the ember-try changes from what I can tell. But, now there are a bunch of eslint failures from within the `tests` directory due to the symlinking of node_modules...

package.json
```diff
@@ -18,7 +18,7 @@
     "test:windows": "ember try:one %EMBER_TRY_SCENARIO% test --skip-cleanup",
     "test:node:dev": "BUILD_DEV=true testem -f testem-node.js",
     "test:null": "echo 'no appropriate changes detected, not running tests'",
-    "prepublish": "node ./bin/install-test-addons.js"
+    "postinstall": "node ./bin/install-test-addons.js"
```
rwjblue: I'm unsure about this. `postinstall` runs for all consumers also (we picked `prepublish` originally since it only runs when you `npm install` in this repo, not `npm install ember-engines`).

Maybe with your other changes, this isn't needed?

trentmwillis: Ah, I knew there was a reason we had used prepublish originally. I just couldn't remember what it was.

I think you're right though, that this shouldn't be needed after the other changes. Will remove. Hopefully, it'll be ready to land after that.

trentmwillis:
Everything is passing now! 🎉 

@rwjblue a review from you before merging would be much appreciated :) then we can land https://github.com/ember-engines/ember-engines/pull/492 which will unblock my fastboot integration work

Issue #404, opened 2017-05-04
Make it possible to exclude files from bundle


mydea:
In one of our engines, we use an addon which loads dependencies on-demand. So basically, there are a bunch of really big and very specific JavaScript dependencies which are manually loaded when entering a route. 

These are JavaScript files that are placed in `/assets/xxx.js` by the addon, which works fine. However, these files are also automatically added to the engine bundle and therefore immediately loaded when entering the engine.

It would be great if there was a way to specify files that should be excluded from the bundle. e.g.:

```js
const Eng = Engine.extend({
  modulePrefix,
  Resolver,
  dependencies: {},
  excludeFromBundle: [
    'assets/file1.js',
    'assets/file2.js'
  ]
});

**Comment by shokmaster** [comment]
+1

**Comment by trentmwillis** [comment]
This is likely a reasonable feature but will need to be added to `ember-asset-loader` itself as well.

**Comment by josex2r** [comment]
Hi @trentmwillis, why is needed to modify `ember-asset-loader`?
The "tree" is built in this addon so I think if its possible to filter the broccoli `Funnel` the files are not going to be included.

For example, something like this will exclude the files located in `addon/` that matchs `**/*-test.js` (_just after [lib/engine-addon.js#L181](https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L181)_)
```javascript
engineSourceTree = new Funnel(engineSourceTree, {
  exclude: ['**/*-test.js'] // this array should be the engine config ("excludeFromBundle") 
});
```

beatrizdemiguelperez:
+1

trentmwillis:
This [feature has now been added to ember-asset-loader](https://github.com/ember-engines/ember-asset-loader/pull/58), but will need to be wired up for use in ember-engines (not entirely sure what that looks like currently).

beatrizdemiguelperez:
hi @mydea, @trentmwillis thinks that https://github.com/ember-engines/ember-engines/pull/497
does not resolve what you are asking for. This PR removes some files from the build engine.js. 
In our project, we need to exclude some files that are only for development purpose. For example, we have example routes, some fixtures/debug/test scripts and we don't want them on production build. This way, we could have more flexibility on final builds (dev, test or production) and configure those builds. What do you think?

mydea:
[...]

PR #497, opened 2017-10-13
#404 exclude files from js bundle

beatrizdemiguelperez:
Add jsBundle config to specify files that should be excluded from the bundle.
Configurable by environment:

Example:
```
var EngineAddon = require('ember-engines/lib/engine-addon');
var debugFiles = /(debug|fixture|\-test)/;
var prodExclude = [debugFiles, 'my-super-addon/routes/examples/*'];

	name: 'my-super-addon',
	lazyLoading: true,
	isDevelopingAddon: function() {
		return true;
	},
	jsFunnel: {
		exclude: {
			production: prodExclude,
			development: debugFiles
		}
	}
});

```

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -35,6 +35,7 @@ var DEFAULT_CONFIG = {
     addon: 'addon',
   },
   minifyCSS: {},
+  jsFunnel: { exclude: {} },
 };
 
 var DEFAULT_BABEL_CONFIG = {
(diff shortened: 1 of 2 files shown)


shokmaster:
Finally the solution is very simply:

```js
// engine/index.js:

var pkg = require('./package');

  name: pkg.name,

  lazyLoading: true,
  ...
  addonJsFiles(tree) {
    return new Funnel(tree, {
      exclude: [/* paths to exclude*/],
      overwrite: true,
      destDir: `${pkg.name}/`
    });
  }
});
```

In our case, we've a pod structure and the test files in the same folder of the .js code, and we need to exclude those test files of the final build.

@beatrizdemiguelperez can you close the PR? Thanks!

PR #494, opened 2017-10-07
Lazy Engines + FastBoot

trentmwillis:
Hopefully the last of these PRs. Prior work can be seen in https://github.com/ember-engines/ember-engines/pull/418 and https://github.com/ember-engines/ember-engines/pull/444.

Todo:
- [x] Fix any test failures (likely to be some on the build side)
- [x] Add tests for new code paths
- [x] Test with ember-engines.com + fastboot

PR #501, opened 2017-11-07
for config, first look in modules, then fallback to DOM

stefanpenner:
needs:

- [ ] properly to be made to work (but conceptually this should work like this)
- [ ] tests

### DIFF
--- a/lib/engine-config-from-meta.js
+++ b/lib/engine-config-from-meta.js
@@ -2,8 +2,13 @@ var config;
 
 try {
   var metaName = '{{MODULE_PREFIX}}/config/environment';
-  var rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
-  config = JSON.parse(unescape(rawConfig));
+  var config;
+  if (metaName in require.entries) {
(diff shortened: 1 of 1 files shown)


rwjblue:
Conceptually, I like the approach in https://github.com/ember-engines/ember-engines/pull/494 better for FastBoot support.

stefanpenner:
> Conceptually, I like the approach in #494 better for FastBoot support.

I'll take a look!

rwwagner90:
@rwjblue @stefanpenner is this PR also for fastboot support, like #494 ? I tried #494 in my app and it threw errors because `document` was not defined, because we were in fastboot and it was trying to grab the config from `document`. If there is anything I can do to help test or further along either PR, let me know! I would love to get engines working with fastboot.

stefanpenner:
Yup good call, closing this in-favor of: https://github.com/ember-engines/ember-engines/pull/494

rwwagner90:
@stefanpenner #494 doesn't work for me either, unfortunately :(. Issues with trying to load configs or something from `document` and `document` doesn't exist in fastboot, so it blows up.

PR #505, opened 2017-12-01
sync blueprint

kellyselden:
with the standalone version https://github.com/ember-engines/ember-engines/blob/master/blueprints/in-repo-engine/routable-files/lib/__name__/config/environment.js


stefanpenner:
thanks @kellyselden

PR #509, opened 2017-12-05
[fix #508] restore previous optimization

### LINKED ISSUE #508: avoidable funnels + merges
https://github.com/ember-engines/ember-engines/blob/8f66b5e3b8089cd53884be49f270cac05f9a3d17/lib/engine-addon.js#L319

is often truthy, when it really isn't needed as `maybeMergeTrees` via `ember-cli/broccoli/merge-trees` will return EMPTY_MERGE_TREES. In this location this internal construct leaks out, and results an un-needed funnel.

This optimization "EMPTY_MERGE_TREES" should likely be made to not leak as it does, but likely a short-term fix is to make `maybeMergeTrees` aware of it, so that downstream code can be made aware.

- [ ] fix
- [ ] see impact (should often-times remove 1 funnel and 1 merge per engine)
- [ ] @rwjblue thoughts?

stefanpenner:
- [ ] confirm in big app
- [ ] tests + related cleanup

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -297,7 +297,12 @@ var buildCompleteJSTree = memoize(function buildCompleteJSTree() {
 function maybeMergeTrees(_inputTrees, options) {
   var inputTrees = _inputTrees.filter(Boolean);
   if (inputTrees.length > 1) {
-    return mergeTrees(inputTrees, options);
+    let merged = mergeTrees(inputTrees, options);
+    if (Array.isArray(merged._inputTrees) && merged._inputTrees.length === 0) {
+      // return undefined, as `mergeTrees` here is returning a merge of nothing
+      return;
+    }
+    return merged;
   } else {
     return inputTrees[0];
   }


lib/engine-addon.js
```diff
@@ -297,7 +297,12 @@ var buildCompleteJSTree = memoize(function buildCompleteJSTree() {
 function maybeMergeTrees(_inputTrees, options) {
   var inputTrees = _inputTrees.filter(Boolean);
   if (inputTrees.length > 1) {
-    return mergeTrees(inputTrees, options);
+    let merged = mergeTrees(inputTrees, options);
+    if (Array.isArray(merged._inputTrees) && merged._inputTrees.length === 0) {
```
rwjblue: This is odd, how can we ever get here?

stefanpenner: mergeTrees is the special mergeTrees you wrote, which is EMPTY_TREE aware.

villander:
@rwjblue @stefanpenner this optimization on `maybeMergeTrees` is ready? We just need to fix the tests?

stefanpenner:
@villander ya, I believe so. The above two checklist items should be accurate.

villander:
Ok @stefanpenner I'll help as soon as I can.

stefanpenner:
> Ok @stefanpenner I'll help as soon as I can.

Awesome would love a hand, I keep getting pulled away before I can work on ^. So your welcome is much appreciated :)

lib/engine-addon.js
villander: @rwjblue any idea of test this function/flow? I don't can export this function to test only.

PR #511, opened 2017-12-20
Add docs about how to test standalone engines

mydea:
While there is a section describing how to do unit/integration testing in in-repo engines, there is no information about how to test standalone engines. Currently, this information is spread out across multiple issues etc. This PR adds a section with basic information on how to test engines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,26 @@ test('renders text', function(assert) {
 
 **Note: you could create a helper and then use it like `Resolver from ../helpers/appointments-manager/resolver`**
 
+### Testing for standalone engines
+
+If you have a lazy engine, you'll need to edit your `tests/test-helper.js` like this:
+
+```js
+import resolver from './helpers/resolver';
+import {
+  setResolver
+} from 'ember-qunit';
+import preloadAssets from 'ember-asset-loader/test-support/preload-assets';
+import manifest from 'dummy/config/asset-manifest';
+import { start } from 'ember-cli-qunit';
+
+setResolver(resolver);
+preloadAssets(manifest).then(start); // This ensures all engine resources are loaded before the tests
+```
+
+This should be enough to make integration & acceptance tests work.
+For unit tests, you'll need to use a custom resolver, as described in [Unit/Integration testing for standalone engines](#unitintegration-testing-for-in-repo-engines).
+
 ## Demo Projects
 
 * [ember-engines-demo](https://github.com/dgeb/ember-engines-demo) - an example of a parent application (consumer).


stefanpenner:
@trentmwillis / @nathanhammond can you guys do a quick review, i want to make sure this is as designed

README.md
```diff
@@ -596,6 +596,27 @@ test('renders text', function(assert) {
 
 **Note: you could create a helper and then use it like `Resolver from ../helpers/appointments-manager/resolver`**
 
+### Testing for standalone engines
+
+If you have an async engine, you'll need to edit your `tests/test-helper.js` like this:
```
trentmwillis: `async` -> `lazy` for consistency with the rest of the documentation.

README.md
```diff
@@ -596,6 +596,27 @@ test('renders text', function(assert) {
...
+import {
+  setResolver
+} from 'ember-qunit';
+import preloadAssets from 'ember-asset-loader/test-support/preload-assets';
+import manifest from 'dummy/config/asset-manifest';
+import { start } from 'ember-cli-qunit';
+
+preloadAssets(manifest); // This ensures all engine resources are loaded before the tests
+setResolver(resolver);
+start();
```
trentmwillis: `preloadAssets` is an asynchronous function that returns a Promise. So this should really be:

```js
preloadAssets(manifest).then(start);
```

Othewise, it is possible that `start()` will start the test suite before all the assets have actually loaded.

README.md
```diff
@@ -596,6 +596,27 @@ test('renders text', function(assert) {
...
+```
+
+This should be enough to make integration & acceptance tests work.
+For unit tests, you'll need to use a custom resolver, as described in [Unit/Integration testing for standalone engines](#unitintegration-testing-for-in-repo-engines).
```
trentmwillis: Since this is a standalone Engine, you should be able to replace the global resolver (set with `setResolver`) to just use the Engine resolver instead, and then you avoid needing a custom resolver for all your unit tests.

mydea: I'm not sure I understand this.

In my standalone, lazy loading engine, I had the following setup:

```js
// tests/test-helper.js
import resolver from './helpers/resolver';
import {
  setResolver
} from 'ember-qunit';
import preloadAssets from 'ember-asset-loader/test-support/preload-assets';
import manifest from 'dummy/config/asset-manifest';
import { start } from 'ember-cli-qunit';

setResolver(resolver);
preloadAssets(manifest).then(start);
```

```js
// tests/helpers/resolver.js
import Resolver from '../../resolver';
import config from '../../config/environment';

const resolver = Resolver.create();

resolver.namespace = {
  modulePrefix: config.modulePrefix,
  podModulePrefix: config.podModulePrefix
};

export default resolver;
```

And that didn't work for me, in my unit tests. (All other tests, e.g. integration & acceptance, work). What would I need to change here?

stefanpenner:
@trentmwillis looks like you suggestions have been addressed, do you mind confirming?

PR #515, opened 2018-01-17
upgrade

stefanpenner:
* nearly that of ember-cli@master blueprint
* chrome over phantom
* extended test matrix
* ember-cli + other safe deps
* all deps
* remove `existsSync` -> `fs.existsSync` ended up not getting deprecated


stefanpenner:
@rwjblue this is finally ready for review.

.travis.yml
```diff
@@ -5,29 +5,43 @@ node_js:
   # so that your addon works for all apps
   - "4"
 
+sudo: false
```
scalvert: Does this not need to be `sudo: required`? I've had to do this a few times lately due to the recent Meltdown/Spectre changes in travis.

.eslintrc.js
```diff
@@ -1,20 +1,48 @@
...
   root: true,
-  extends: ['eslint:recommended', 'prettier'],
-  plugins: ['prettier'],
   parserOptions: {
     ecmaVersion: 2017,
+    sourceType: 'module'
   },
+  plugins: [
+    'ember'
+  ],
```
scalvert: Prettier is added as a devDep, but isn't configured. Did you intend to configure it?

tests/dummy/lib/ember-blog/addon/components/hello-name.js
```diff
@@ -1,16 +1,21 @@
...
+    this._later = later(() => {
       if (this.isDestroyed || this.isDestroying) {
         return;
       }
       this.set('name', 'Jerry');
-    }, 500);
+    }, 50);
   },
+  destroy() {
+    cancel(this._later);
```
scalvert: Good catch.

.eslintrc.js
stefanpenner: We can configure it in a later PR.

.travis.yml
stefanpenner: fixed

node-tests/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
rwjblue: We should be able to remove this file by adding an overrides config that matches `node-tests/**` with these settings to the top level `.eslintrc.js` file...

tests/dummy/lib/ember-chat/addon/views/application.js
```diff
@@ -1,20 +1,5 @@
-import Ember from 'ember';
+import Component from '@ember/component';
```
rwjblue: I think we can remove this file, we never look up `views:application` as of Ember 2.x...

PR #529, opened 2018-03-26
fix: ensure the "environment" is passed to the engine's config generator

lennyburdette:
cc @trentmwillis 

The latest version of ember-engines is breaking an expectation in our apps that the "environment" key will always be set in `{{modulePrefix}}/config/environment`.

This change always grabs the `env` from the host application instead of assuming that it will be passed in from hooks like `contentFor`. @trentmwillis memoized the `engineConfig` method in the fastboot updates, and passing arguments to a memoized method can be problematic.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -976,14 +976,16 @@ module.exports = {
 
       @public
       @method engineConfig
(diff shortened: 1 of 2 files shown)


rwjblue:
Thanks @lennyburdette!

PR #532, opened 2018-03-31
Fix funnel + merges

### LINKED ISSUE #508: avoidable funnels + merges

villander:
It's fix #508 and replace the PR #509 

- [x] Tests

- [x]  Confirm in big app

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -13,6 +13,7 @@ var defaultsDeep = require('lodash/defaultsDeep');
 var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 var Addon = require('ember-cli/lib/models/addon');
 var memoize = require('./utils/memoize');
+var maybeMergeTrees = require('./utils/maybe-merge-trees');
 var p = require('ember-cli-preprocess-registry/preprocessors');
 var CompactReexports = require('babel-plugin-compact-reexports');
 var shouldCompactReexports = require('./utils/should-compact-reexports');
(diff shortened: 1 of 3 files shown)


stefanpenner:
I can also likely test this weekend.

villander:
@stefanpenner How will you test this? What is the scenario?

stefanpenner:
@villander I was planning on doing it 2 ways:

* confirm by stepping through the code, that it is working as expected
* confirm by running against once of our large apps (that has like 40 engines or so)

Obviously you can't test in our app, but I think any engine app you have access to should be sufficient.

stefanpenner:
I feel pretty good about this PR, so as long as you have also tested it and confirmed the behavior, we can move forward :)

villander:
Yes @stefanpenner make sense for me.

In addition to the unit tests, I've already tested this with ember-engines website (which is using Fastboot as well). So I think we can move forward.

stefanpenner:
awesome, thanks!

PR #538, opened 2018-04-07
Fix regression environment

villander:
cc: @kellyselden @rwjblue

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1,30 +1,29 @@
 'use strict';
 
-var Funnel = require('broccoli-funnel');
-var merge = require('lodash/merge');
-var mergeTrees = require('ember-cli/lib/broccoli/merge-trees');
-var fs = require('fs');
-var path = require('path');
-var writeFile = require('broccoli-file-creator');
-var Babel = require('broccoli-babel-transpiler');
-var concat = require('broccoli-concat');
-var DependencyFunnel = require('broccoli-dependency-funnel');
-var defaultsDeep = require('lodash/defaultsDeep');
-var calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
-var Addon = require('ember-cli/lib/models/addon');
-var memoize = require('./utils/memoize');
-var maybeMergeTrees = require('./utils/maybe-merge-trees');
-var p = require('ember-cli-preprocess-registry/preprocessors');
-var CompactReexports = require('babel-plugin-compact-reexports');
-var shouldCompactReexports = require('./utils/should-compact-reexports');
-var preprocessCss = p.preprocessCss;
-var preprocessMinifyCss = p.preprocessMinifyCss;
-var BroccoliDebug = require('broccoli-debug');
+const Funnel = require('broccoli-funnel');
+const merge = require('lodash/merge');
+const mergeTrees = require('ember-cli/lib/broccoli/merge-trees');
+const fs = require('fs');
+const path = require('path');
+const writeFile = require('broccoli-file-creator');
+const concat = require('broccoli-concat');
+const DependencyFunnel = require('broccoli-dependency-funnel');
+const defaultsDeep = require('lodash/defaultsDeep');
+const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
+const Addon = require('ember-cli/lib/models/addon');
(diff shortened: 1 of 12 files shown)


mike183:
@lennyburdette: You may also want to check this doesn't break the work you did in #529

lennyburdette:
@mike183 I think we can go with this change instead of mine. They accomplish the same thing.

lennyburdette:
Oh wait, mine was merged already. I’ll check that it doesn’t break our app again. Thanks for the heads up!

villander:
@lennyburdette your solution had a hole, where the value of environment, was always `development` or `undefined`. Now it's all right, I imagine. Test your app and let me know, please.

node-tests/acceptance/build-test.js
```diff
@@ -73,10 +73,7 @@ describe('Acceptance', function() {
         expect(output.manifest()).to.deep.equal(expectedManifests['lazy']);
         output.contains('assets/node-asset-manifest.js');
         // Engine inherites its environment from the parent
-        output.contains(
-          `engines-dist/${engineName}/config/environment.js`,
-          /"environment":"development"/
```
rwjblue: Why would we want this assertion removed?

villander: @rwjblue the old solution has a problem where the `environment` always returns the` development` value. And that test just covered that scenario, I preferred to make sure that only the `config/environment` is exposed out.

But I agree we should test all scenarios.

villander:
> We need to change the way we cache here, specifically we need to support building two environments during a single build (this is how ember s + /tests actually works.

Can you tweak the caching to a map (or other structure that supports different configs per-environment)?

@rwjblue you can show me some example of how `ember-cli` do this?

rwjblue:
[Here](https://github.com/ember-cli/ember-cli/blob/ea4e7758c37419a257d317c37581f901d8e63ced/lib/models/project.js#L222-L257) is where ember-cli manages the config map.

stefanpenner:
I believe @rwjblue  is recommending something like:

```
diff --git a/lib/engine-addon.js b/lib/engine-addon.js
index c15e051..faf8bf5 100644
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -964,13 +964,19 @@ module.exports = {
       @return {Object} Configuration object that will be provided to the engine.
     */
     options.engineConfig = function() {
-      if (this._engineConfig) {
-        return this._engineConfig;
-      }
-
       var host = findHost.call(this);
       var env = host.env;
+      let key = `${host.project.configPath()}|${env}`;
+      if (!this._engineConfig) {
+        this._engineConfig = new Map();
+      }
+
+      if (this._engineConfig.has(key)) {
+        this._engineConfig.get(key);
+      }
+
       var baseConfig = {};
+      var config;
 
       var configPath = 'config';
 
@@ -990,10 +996,13 @@ module.exports = {
 
         var addonsConfig = this.getAddonsConfig(env, engineConfig);
 
-        return (this._engineConfig = merge(addonsConfig, engineConfig));
+        config = merge(addonsConfig, engineConfig);
       } else {
-        return (this._engineConfig = this.getAddonsConfig(env, {}));
+        config =  this.getAddonsConfig(env, {});
       }
+
+      this._engineConfig.set(key, config);
+      return config;
     };
 
     /**
@@ -1166,6 +1175,7 @@ module.exports = {
 
       return mergedTree;
     };
+
     return options;
   },
 };

```

node-tests/acceptance/build-test.js
stefanpenner: yes, we will need to test more then one scenario

villander: I'll do this between today and tomorrow.

stefanpenner:
crap.... i pushed the wrong branch when I tried to rebase....

stefanpenner:
since i screwed this up, I'll take it to completion: new branch/pr -> https://github.com/ember-engines/ember-engines/pull/545

villander:
@stefanpenner I was working on this now haha

I'll leave you finish this =D

stefanpenner:
@villander oh geeze, I am sorry. I attempted to merely help with the rebase, but I managed to fat finger this one..

villander:
No problem @stefanpenner. I'll learn from you =D

Issue #549, opened 2018-05-09
0.5.19: "Defining a custom serialize method on an Engine route is not supported."


lennyburdette:
This change does not appear to work: https://github.com/ember-engines/ember-engines/commit/6c598e62b5bb176562631a23fbe87bfb88cefe5e#diff-053886e73d8fe44d16cce9b6dbe79118R10

When I throw a debugger in that module and eval `Ember.Route.prototype.serialize`, I get `undefined`.

It could be `Ember.Route.create().serialize`, but that seems mildly inefficient. @rwjblue any other ideas?

CodeOfficer:
I'm seeing this as well. Same code works fine with 0.5.18 but breaks with this error in 0.5.19.

rwjblue:
:( sorry y’all

Fix should be to call `Ember.Route.proto().serialize` and cache it I believe.

lennyburdette:
It also needs a `return` in front 🤣. But that fixes it!

rwjblue:
ember-engines 0.5.20 published 🎉

PR #550, opened 2018-05-09
Fix reference to Route's defaultSerialize

lennyburdette:
Addresses #549


rwjblue:
Thank you @lennyburdette!

PR #553, opened 2018-05-15
add public API for reseting the `memoize` cache

lennyburdette:
I encountered the need for this change when I tried writing a test for
an addon that interacts with ember-engines. The cache holds on to
broccoli trees, including their location on disk. I was using
[broccoli-test-helper][bth] which creates temp directories for each
test. Ember-engines would continue to refer to the previous test's
temp directories and fail in confusing ways. This new API allows me
to reset the caches between tests.

[bth]:https://github.com/broccolijs/broccoli-test-helper

@rwjblue

### DIFF
--- a/addon/-private/link-to-utils.js
+++ b/addon/-private/link-to-utils.js
@@ -1,8 +1,8 @@
-import Ember from 'ember';
+import { VERSION } from '@ember/version'
 
 // LAME, but ¯\_(ツ)_/¯
 export default function hasEmberVersion(major, minor) {
-  var numbers = Ember.VERSION.split('-')[0].split('.');
+  var numbers = VERSION.split('-')[0].split('.');
   var actualMajor = parseInt(numbers[0], 10);
   var actualMinor = parseInt(numbers[1], 10);
   return actualMajor > major || (actualMajor === major && actualMinor >= minor);
--- a/lib/test-utils.js
+++ b/lib/test-utils.js
@@ -1,0 +1,9 @@
+const resetMemoize = require('./utils/memoize')._reset;
+
+/**
+ * Reset cached state (useful in tests).
+ * @public
+ */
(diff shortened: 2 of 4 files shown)


lennyburdette:
@rwjblue sounds good ... would you test the utils/memoize module in isolation or a more integration-style test that calls the memoized functions in lib/engine-addon?

rwjblue:
I'd suggest (for now) just a unit test of the memoize function (independent of lib/engine-addon) is a good start.

I believe that @stefanpenner and I would like to migrate the tests to the `broccoli-test-helper` style (that you mentioned in your discourse post), and we can/will add more acceptancey coverage for this then...

lennyburdette:
Unit test added!

lennyburdette:
@rwjblue would you mind restarting the "EMBER_TRY_SCENARIO=ember-beta" travis build? Looks like a transient issue.

rwjblue:
restarted!

lennyburdette:
thanks, it passed!

stefanpenner:
rather then a reset, doesn't it seem like the memoization cache key isn't correct?

lennyburdette:
@stefanpenner Hmm that's an interesting idea. I'm trying to run
 
```js
const output = broccoliTestHelper.createBuilder(new EmberApp.toTree());
output.build();
```

in separate tests in the same process. I could imagine updating memoize() to take a unique key per build, but I'm not sure how to pass something from EmberApp down to the internals of ember-engine.

I'm open to ideas! It'd be great if memoization Just Worked™️ instead of requiring special cleanup.

rwjblue:
one idea is to include some an addon instance uuid in `init` and then include that in the cache key?

lennyburdette:
Yeah, I think that'll work! PR incoming.

lennyburdette:
closing in favor of #555

PR #554, opened 2018-05-16
upgrade ember-cli

stefanpenner:
Doing this incrementally, commit at a time. To get travis to also run the full test suite after each step. Review is likely not required until complete.

- [x] complete


rwjblue:
This seems good, but I don’t see an ember-cli update in the diff...

villander:
@stefanpenner you already make this here right?

https://github.com/ember-engines/ember-engines/commit/35e7a901a3f558ef9f0eb47705aa9bc11265c6f7

stefanpenner:
@villander sorta, the goal of this one is to bump to the beta of ember-cli, as we are experience issues with engines and the beta.

stefanpenner:
now to wait patiently as CI completes...

stefanpenner:
> Quick .travis.yml experiment, should likely switch to jobs/stages

stefanpenner:
not perfect, but https://travis-ci.org/ember-engines/ember-engines/builds/385973906 makes actually seeing whats failing on CI much more reasonable. Also now we actually test stuff..

stefanpenner:
It looks like if a UI test fails, and its time's out each test will add 60s of overhead, currently resulting in 16m+ of time waste.

rwjblue:
Agreed, lets remove ember-lts-2.16 from the test matrix.

RE: beta, there is at least one "known issue" is [this require](https://github.com/ember-engines/ember-engines/blob/864d391bebb7e658d93580c11257130475b47aa7/tests/unit/engine-instance-test.js#L11) needs to be updated (the path inside ember's internals changed

stefanpenner:
@rwjblue I'll start working on the node failing tests. It it safe to assume `beta` issue will get sorted out without my intervention?

stefanpenner:
Only NODE_JOB remains 🕺

stefanpenner:
I think a valid alternative is to do:

```js
  // We check the local ember-cli-babel version and the global ember-cli and
  // loader.js versions
  return (
    checker.for('ember-cli-babel', 'npm').gte('6.0.0') &&
    checker.for('ember-cli', 'npm').gte('2.13.0') &&
    checker.for('loader.js', 'npm').gte('4.4.0')
  );
```

stefanpenner:
its GREEN!

villander:
Great stuff @stefanpenner 

thanks for the hard work

PR #565, opened 2018-07-12
fix: find-hosts-host 'Cannot read property 'enabled' of undefined'

knownasilya:
Encountered this error with fresh install and in-repo-engine with lazyloading enabled. 

```
> ember serve

Could not start watchman
Visit https://ember-cli.com/user-guide/#watchman for more info.
Cannot read property 'enabled' of undefined
```

This error might be happening because I have another in-repo-addon.

### DIFF
--- a/lib/utils/find-hosts-host.js
+++ b/lib/utils/find-hosts-host.js
@@ -8,7 +8,7 @@ module.exports = function findHostsHost() {
   // Keep iterating upward until we don't have a grandparent.
   // Has to do this grandparent check because at some point we hit the project.
   do {
-    if (current.lazyLoading.enabled === true) {
+    if (current && current.lazyLoading && current.lazyLoading.enabled === true) {
       if (foundHost) {
         return current;
       }


rwjblue:
Hmm, I thought we “upconverted” a Boolean value to the POJO before this was called?

knownasilya:
I just edited my initial description regarding why I think it happens.

knownasilya:
Repro can be found here https://github.com/knownasilya/ember-ghost/commit/40b131e035e6f414fdf61860de2f34b0686e20c7 (note the app doesn't start once the fix is applied, that is unrelated).

rwjblue:
Yes, exactly

lib/utils/find-hosts-host.js
```diff
@@ -8,7 +8,7 @@ module.exports = function findHostsHost() {
   // Keep iterating upward until we don't have a grandparent.
   // Has to do this grandparent check because at some point we hit the project.
   do {
-    if (current.lazyLoading.enabled === true) {
+    if (current && current.lazyLoading && current.lazyLoading.enabled === true) {
```
villander: @knownasilya what scenario `current.lazyLoading` is undefined? 

can you add tests for this?

villander:
It's already done by #610, can you close @rwjblue please?

PR #571, opened 2018-07-24
Start fixing parallel builds

stefanpenner:
This will depend on ember-cli + ember-data releases (and maybe more):

depends on:

* https://github.com/emberjs/data/pull/5535
* https://github.com/ember-cli/ember-cli/pull/7934
* https://github.com/babel/broccoli-babel-transpiler/pull/146
* maybe more


rwjblue:
I think all of the internals needed for this are landed, and updated. I've rebased this to see / confirm that we can land it now...

rwjblue:
Hmm, looks like we need more work I guess (though I'm surprised).

PR #583, opened 2018-08-09
fix: Use the tree passed to treeForAddon if provided.

amiller-gh:
Brings engines closer in line with the behavior of regular addons. 

Some addons (including the `ember-engines` addon itself!) rely on the ability to override parent addons' `treeFor*` hooks and intercept trees before they are processed by the parent. Ex:

```javascript
// ... 
included(parent){
  let oldHook = parent.treeForAddon;
  parent.treeForAddon = (tree) => {
    tree = new DoSomething(tree);
    return oldHook.call(parent, tree);
  }
}
```

Ember engines currently throws out any tree passed to `treeForAddon` and creates it fresh from disk, breaking this convention.

I'm not entirely sure how to test this. Suggestions welcome. 

Also, this fix also only applies to eager engines – lazy engines are almost entirely implemented in the `treeForPublic` hook (why is that again?) and I believe will take a lot more work to enable this for.

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -84,10 +84,10 @@ const buildVendorCSSTree = memoize(function buildVendorCSSTree(vendorTree) {
   return this.debugTree(vendorCSSTree, 'vendor-style:input');
 });
 
-const buildEngineJSTree = memoize(function buildEngineJSTree() {
+const buildEngineJSTree = memoize(function buildEngineJSTree(tree) {
   let engineSourceTree;
   let treePath;
-  let addonTree = this.options.trees.addon;
+  let addonTree = tree || this.options.trees.addon;
 
   if (typeof addonTree === 'string') {
     treePath = path.resolve(this.root, addonTree);
@@ -120,8 +120,8 @@ const buildEngineJSTree = memoize(function buildEngineJSTree() {
 });
 
 const buildEngineJSTreeWithoutRoutes = memoize(
-  function buildEngineJSTreeWithoutRoutes() {
-    return new DependencyFunnel(buildEngineJSTree.call(this), {
+  function buildEngineJSTreeWithoutRoutes(tree) {
+    return new DependencyFunnel(buildEngineJSTree.call(this, tree), {
       exclude: true,
       entry: this.name + '/routes.js',
       external: ['ember-engines/routes'],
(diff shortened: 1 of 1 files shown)


rwjblue:
Haven't dug in completely, seems somewhat related to https://github.com/ember-engines/ember-engines/pull/428. Does this supercede that effort?

amiller-gh:
Ah good question! They do look related, but if i'm understanding it correctly this is in *addition* to it. 

#428 calls the old `treeForAddon`, this enables other addons to override an engine's `treeForAddon` and intercept the addon tree before the engine is built, like what we have to do to [enable CSS Blocks in Ember CLI](https://github.com/linkedin/css-blocks/pull/185/files#diff-448db4de6323e80ac7e1aea43a4aba07R186). 

I was able to [hack around it](https://github.com/linkedin/css-blocks/pull/185/files#diff-448db4de6323e80ac7e1aea43a4aba07R300), but technically ember-engines is going against convention here by entirely throwing out the tree provided to it by the `treeForAddon` hook and starting fresh.

@stefanpenner who also has context here.

amiller-gh:
Updated main description with simple statement of problem.

lib/engine-addon.js
```diff
@@ -84,10 +84,10 @@ const buildVendorCSSTree = memoize(function buildVendorCSSTree(vendorTree) {
   return this.debugTree(vendorCSSTree, 'vendor-style:input');
 });
 
-const buildEngineJSTree = memoize(function buildEngineJSTree() {
+const buildEngineJSTree = memoize(function buildEngineJSTree(tree) {
   let engineSourceTree;
   let treePath;
-  let addonTree = this.options.trees.addon;
+  let addonTree = tree || this.options.trees.addon;
```
stefanpenner: Now that `tree` is passed in, under what circumstances should we continue to consult `this.options.trees.addon` ?

My intuition (i'll need to verify) is that your change, removes the need to invoke consult `this.options.trees.addon` at all. Could you investigate?

stefanpenner:
I believe this is a good path. I would like to get the next release out first, then experiment with this change in a production app, make sure all seems well.

lib/engine-addon.js
amiller-gh: I think that is correct – I kept this fallback in to ensure full back compat. Will test it out a bit more and get back to you.

amiller-gh:
@stefanpenner, what can I do to help you push this forward?

villander:
@stefanpenner @rwjblue what is missing here? I can help

PR #593, opened 2018-09-23
Fix router setup

chadhietala:
This needed because we need to infer the setup of the router.js. Prior to https://github.com/emberjs/ember.js/pull/16974 we used to call `_getHandlerFunction` to get the closed over function to resolve a name to a handler. https://github.com/emberjs/ember.js/pull/16974 removed this private method.


addon/-private/router-ext.js
```diff
@@ -51,12 +58,27 @@ Router.reopen({
...
+  },
+
+  _handlerResolver() {
     let seen = this._seenHandlers;
     let owner = getOwner(this);
-
     return name => {
       let engineInfo = this._engineInfoByRoute[name];
-
+      debugger;
```
rwjblue: debugger

stefanpenner:
@chadhietala yuno test?

PR #594, opened 2018-09-27
Exclude fingerprinted environment files

chrisseto:
Enabling ember-cli's fingerprinting adds environment.js to engines' asset-manifest as `filesToIgnore` does not take the fingerprint into account.
This can be worked around be adding `config/environment.js` to the fingerprinting's `exclude` key but may be difficult to track down.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ module.exports = ManifestGenerator.extend({
   name: 'ember-engines',
   manifestOptions: Object.freeze({
     bundlesLocation: 'engines-dist',
-    filesToIgnore: [/\/config\/environment.js$/],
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -7,7 +7,7 @@ module.exports = ManifestGenerator.extend({
   name: 'ember-engines',
   manifestOptions: Object.freeze({
     bundlesLocation: 'engines-dist',
-    filesToIgnore: [/\/config\/environment.js$/],
+    filesToIgnore: [/\/config\/environment(?:-[\w\d]+)?\.js$/],
```
villander: @chrisseto can you add tests for this, please?

villander:
@dgeb this PR make sense?

Issue #405, opened 2017-05-04
Engines' dependency addons are `include`d more than once


dfreeman:
We first came across this with [`ember-get-config`](https://github.com/null-null-null/ember-get-config), which [emits a warning](https://github.com/null-null-null/ember-get-config/blob/master/index.js#L38) if a single instance has its `included()` hook invoked more than once. On investigation, it seems that any addon in an engine's `dependencies` gets included twice for the engine.

A simple reproduction is available [here](https://github.com/dfreeman/engine-get-config-warning). If you log or set a breakpoint, `ember-get-config`'s `included()` hook is invoked three times: once for the dummy app, and twice for the engine (which triggers the warning message).

buschtoens:
I can confirm that this is still an issue with the latest version:

- `ember-engines`: `0.8.19`
- `ember-cli`: `3.28.2`
- `ember-source`: `3.28.1`

I am in the process of upgrading the main app of @ClarkSource's monorepo to use `ember-auto-import@2`, but kept failing with Babel complaining about the `ember-cli-htmlbars:inline-precompile` being duplicated.

<details>
<summary>Error Log</summary>

```
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-maybe-in-element]) in ember-maybe-in-element/components/maybe-in-element.js

Duplicate plugin/preset detected.
If you'd like to use two separate instances of a plugin,
they need separate names, e.g.

  plugins: [
    ['some-plugin', {}],
    ['some-plugin', {}, 'some unique name'],
  ]

Duplicates detected are:
[
  {
    "name": "ember-cli-htmlbars:inline-precompile",
    "alias": "/Users/jan/clark/application/node_modules/babel-plugin-ember-template-compilation/src/node-main.js",
    "options": {
      "enableLegacyModules": [
        "ember-cli-htmlbars",
        "ember-cli-htmlbars-inline-precompile",
        "htmlbars-inline-precompile"
      ]
    },
    "dirname": "/Users/jan/clark/application/client",
    "ownPass": false,
    "file": {
      "request": "/Users/jan/clark/application/node_modules/babel-plugin-ember-template-compilation/src/node-main.js",
      "resolved": "/Users/jan/clark/application/node_modules/babel-plugin-ember-template-compilation/src/node-main.js"
    }
  },
  {
    "name": "ember-cli-htmlbars:inline-precompile",
[...]

PR #595, opened 2018-09-28
Deduplicate addons in lazy engines that are nested dependencies of the engine

2hu12:
Try to fix remaining "Shared addon deeper than engine" problem of https://github.com/ember-engines/ember-engines/pull/356.

An example of this problem copied from https://github.com/ember-engines/ember-engines/pull/356#issuecomment-288571816

> - project has dep on addon-derp
> - project has lazy engine engine-a
> - engine-a has engine addon-z
> - addon-z has addon-derp (addon-derp should not be duplicated)

For the example above, the `addon-derp` should not be duplicated, but without this PR it would be duplicated in the `vendor.js` of `engine-a`.

@rwjblue @trentmwillis @stefanpenner

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -249,6 +249,49 @@ const buildEngineStyleTree = memoize(function buildEngineStyleTree() {
   );
 });
 
+const recoverAddons = function recoverAddons(addon) {
+  if (addon._orginalAddons) {
+    addon.addons = addon._orginalAddons;
+  }
+
+  if (addon.addons.length > 0) {
+    addon.addons.forEach(recoverAddons)
+  }
+}
+
+const deeplyNonDuplicatedAddon = function deeplyNonDuplicatedAddon(hostAddons, addon, treeName) {
+  if (addon.addons.length === 0) {
+    return;
+  }
+
+  addon._orginalAddons = addon.addons;
+  addon.addons = addon.addons.filter(innerAddon => {
+    let shouldDeepDedup = !(innerAddon.lazyLoading && innerAddon.lazyLoading.enabled);
+
+    if (shouldDeepDedup && innerAddon.addons.length > 0) {
+      innerAddon._orginalAddons = innerAddon.addons;
+      deeplyNonDuplicatedAddon(hostAddons, innerAddon, treeName);
+    }
+
+    let hostAddon = hostAddons[innerAddon.name];
+
+    if (hostAddon && hostAddon.cacheKeyForTree) {
+      let innerCacheKey = innerAddon.cacheKeyForTree(treeName);
+      let hostAddonCacheKey = hostAddon.cacheKeyForTree(treeName);
+
+      if (
+        innerCacheKey != null &&
+        innerCacheKey === hostAddonCacheKey
+      ) {
+        // the addon specifies cache key and it is the same as host instance of the addon, skip the tree
+        return false;
+      }
+    }
+
+    return true;
+  });
+}
+
 module.exports = {
   extend(options) {
     let originalInit =
@@ -330,36 +373,27 @@ module.exports = {
       let invokeArguments = args || [];
       let hostAddons = this.ancestorHostAddons();
 
-      return this.addons
-        .map(addon => {
+      if (methodName === 'treeFor') {
+        deeplyNonDuplicatedAddon(hostAddons, this, invokeArguments[0]);
+      }
 
+      let trees = this.addons
+        .filter(addon => {
           if (!addon[methodName]) {
             // no method to call
-            return;
+            return false;
           }
-
-          let hostAddon = hostAddons[addon.name];
-
-          if (hostAddon && hostAddon.cacheKeyForTree) {
-            if (methodName === 'treeFor') {
-              let treeName = invokeArguments[0];
-              let hostAddonCacheKey = hostAddon.cacheKeyForTree(treeName);
-              let addonCacheKey = addon.cacheKeyForTree(treeName);
-
-              if (
-                addonCacheKey != null &&
-                addonCacheKey === hostAddonCacheKey
-              ) {
-                // the addon specifies cache key and it is the same as host instance of the addon, skip the tree
-                return;
-              }
-            } else {
-              return;
-            }
-          }
-
+          return true;
+        })
+        .map(addon => {
           return addon[methodName].apply(addon, invokeArguments);
-        }).filter(Boolean);
+        });
+
+      if (methodName === 'treeFor') {
+        recoverAddons(this);
+      }
+
+      return trees;
     };
 
     options.debugTree = BroccoliDebug.buildDebugCallback(
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -14,7 +14,7 @@ const Addon = require('ember-cli/lib/models/addon');
 const memoize = require('./utils/memoize');
 const maybeMergeTrees = require('./utils/maybe-merge-trees');
 const deeplyNonDuplicatedAddon = require('./utils/deeply-non-duplicated-addon');
-const recoverAddons = require('./utils/recover-addons');
+const restoreOriginalAddons = require('./utils/restore-original-addons');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const shouldCompactReexports = require('./utils/should-compact-reexports');
 const appendCompactReexportsIfNeeded = require('./utils/append-compact-reexports-if-needed');
@@ -332,28 +332,65 @@ module.exports = {
 
       let invokeArguments = args || [];
       let hostAddons = this.ancestorHostAddons();
+      let treeName = invokeArguments[0];
 
-      if (methodName === 'treeFor') {
-        deeplyNonDuplicatedAddon(hostAddons, this, invokeArguments[0]);
+      // methodName could be "treeFor" or "included"
+      // the "included" methodName will be handled in the old implementation for now
+      // TODO: when deeplyNonDuplicatedAddon is completely ready
+      // make sure that the "included" methodName is considered
+      if (process.env.EMBER_ENGINES_ADDON_DEDUPE && methodName === 'treeFor') {
+        let trees
+
+        try {
+          deeplyNonDuplicatedAddon(hostAddons, this, treeName);
+
+          trees = this.addons
+            .filter(addon => {
+              if (!addon[methodName]) {
+                // no method to call
+                return false;
+              }
+              return true;
+            })
+            .map(addon => {
+              return addon[methodName].apply(addon, invokeArguments);
+            });
+        } finally {
+          restoreOriginalAddons(this);
+        }
+
+        return trees
       }
 
-      let trees = this.addons
-        .filter(addon => {
+      // old implementation
+      return this.addons
+        .map(addon => {
           if (!addon[methodName]) {
             // no method to call
-            return false;
+            return;
           }
-          return true;
-        })
-        .map(addon => {
+
+          let hostAddon = hostAddons[addon.name];
+
+          if (hostAddon && hostAddon.cacheKeyForTree) {
+            if (methodName === 'treeFor') {
+              let hostAddonCacheKey = hostAddon.cacheKeyForTree(treeName);
+              let addonCacheKey = addon.cacheKeyForTree(treeName);
+
+              if (
+                addonCacheKey != null &&
+                addonCacheKey === hostAddonCacheKey
+              ) {
+                // the addon specifies cache key and it is the same as host instance of the addon, skip the tree
+                return;
+              }
+            } else {
+              return;
+            }
+          }
+
           return addon[methodName].apply(addon, invokeArguments);
-        });
-
-      if (methodName === 'treeFor') {
-        recoverAddons(this);
-      }
-
-      return trees;
+        }).filter(Boolean);
     };
 
     options.debugTree = BroccoliDebug.buildDebugCallback(
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -18,6 +18,7 @@ const recoverAddons = require('./utils/recover-addons');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const shouldCompactReexports = require('./utils/should-compact-reexports');
 const appendCompactReexportsIfNeeded = require('./utils/append-compact-reexports-if-needed');
+const ensureLazyLoadingHash = require('./utils/ensure-lazy-loading-hash');
 const preprocessCss = p.preprocessCss;
 const preprocessMinifyCss = p.preprocessMinifyCss;
 const BroccoliDebug = require('broccoli-debug');
@@ -378,13 +379,7 @@ module.exports = {
       this._engineConfig = new Map();
       this.options = defaultsDeep(options, DEFAULT_CONFIG);
 
-      // Ensure lazyLoading is a hash, retain backwards compatibility with using
-      // a boolean value
-      if (typeof this.lazyLoading === 'boolean') {
-        this.lazyLoading = {
-          enabled: this.lazyLoading,
-        };
-      }
+      ensureLazyLoadingHash(this);
 
       // NOTE: This is a beautiful hack to deal with core object calling toString on the function.
       // It'll throw a deprecation warning if this isn't present because it doesn't see a `_super`
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -13,6 +13,8 @@ const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 const Addon = require('ember-cli/lib/models/addon');
 const memoize = require('./utils/memoize');
 const maybeMergeTrees = require('./utils/maybe-merge-trees');
+const deeplyNonDuplicatedAddon = require('./utils/deeply-non-duplicated-addon');
+const recoverAddons = require('./utils/recover-addons');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const shouldCompactReexports = require('./utils/should-compact-reexports');
 const appendCompactReexportsIfNeeded = require('./utils/append-compact-reexports-if-needed');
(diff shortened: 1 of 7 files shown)


stefanpenner:
@2hu12 very nice, I was hoping to tackle this soon. But I am happy you beat me to it. Let me look over it :)

lib/engine-addon.js
```diff
@@ -245,6 +245,49 @@ const buildEngineStyleTree = memoize(function buildEngineStyleTree() {
...
+  if (addon._orginalAddons) {
+    addon.addons = addon._orginalAddons;
+  }
+
+  if (addon.addons.length > 0) {
+    addon.addons.forEach(recoverAddons)
+  }
+}
+
+const deeplyNonDuplicatedAddon = function deeplyNonDuplicatedAddon(hostAddons, addon, treeName) {
```
stefanpenner: could this function be unit tested? That will make it easier to review as different scenarios can easily be validated :).

2hu12: Add test for "deeplyNonDuplicatedAddon".

stefanpenner:
@2hu12 thanks for adding tests, will check over this !

villander:
@rwjblue @stefanpenner can we merge this guys?

mydea:
Would also love to have this, I tried this branch and it fixed our remaining deduplication issues perfectly.

lib/utils/deeply-non-duplicated-addon.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {String} treeName
+ */
+module.exports = function deeplyNonDuplicatedAddon(hostAddons, dedupedAddon, treeName) {
+  if (dedupedAddon.addons.length === 0) {
+    return;
+  }
+
+  dedupedAddon._orginalAddons = dedupedAddon.addons;
+  dedupedAddon.addons = dedupedAddon.addons.filter(addon => {
+    if (addon.lazyLoading && addon.lazyLoading.enabled) {
```
villander: you need check if `addon.lazyLoading` is a bollean, please

villander:
@2hu12 @stefanpenner we need an optimizing deduplication hook in Ember CLI which has insight across all boundaries with all information pulled out of variables, and then an asset dep tree to make sure that we can successfully merge nodes when their node and their subtrees are identical.

another point: we need to figure out if we want to allow different versions of the same addon to coexist (via namespacing)

@rwjblue @dgeb

2hu12:
> @2hu12 @stefanpenner we need an optimizing deduplication hook in Ember CLI which has insight across all boundaries with all information pulled out of variables, and then an asset dep tree to make sure that we can successfully merge nodes when their node and their subtrees are identical.

I think the addon duplication is an issue when there is a `lazyLoading` option and for now only ember-engine has this option, is that right?

> another point: we need to figure out if we want to allow different versions of the same addon to coexist (via namespacing)

I have a similar question for @rwjblue on [discuss](https://discuss.emberjs.com/t/how-should-we-resolve-conflicting-addons-and-dependencies/14523/10) with the [answer](https://discuss.emberjs.com/t/how-should-we-resolve-conflicting-addons-and-dependencies/14523/11).

villander:
> I think the addon duplication is an issue when there is a lazyLoading option and for now only ember-engine has this option, is that right?

sorry I edited my comment. We make sure that we can successfully merge nodes when their node and their subtrees are identical.

> I have a similar question for @rwjblue on discuss with the answer.

@2hu12  ember-cli is different of ember-engines project, we can have two engines with the same addon only with different versions (via namespacing). I talked about it on last week with @dgeb. The use case is very different. Rob also said that he thinks not, we are still not sure. And as you know more than I dependencies on ember-engines is something even more obscure yet undocumented, it's something I wanted very much to see settled.

lib/utils/deeply-non-duplicated-addon.js
dgeb: we should be able to reuse `ensureLazyLoadingHash` from #610

2hu12: I think https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L384-L390 already makes sure that `addon.lazyLoading` is a hash. Do we still have to use `ensureLazyLoadingHash` each time before we use `addon.lazyLoading`?

@villander Have you got a issue if you don't have [#610](https://github.com/ember-engines/ember-engines/pull/610)?

villander: @2hu12 What Dan wanted to say is that we now have a utility and it makes more sense to keep the code standardized from now. 

And in your code too, if the check in hook init (https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L384-L390) already solves your problem why do the "if" here again?

2hu12: @villander Thank you for the explanation.

The `if` here is to make sure that there's a `lazyLoading` option and the `enabled` property is `true`, if one engine has no `lazyLoading` at all using `if (addon.lazyLoading.enabled)` would cause runtime error.

And my question is should the `ensureLazyLoadingHash` helper be used once in the `init` hook or every time before it reads the `addon.lazyLoading.enabled`?

villander: @2hu12 In your case, it should be used in hook init and in your solution because you are doing a scan on all addons/dependencies

try to use only in init and see if something breaks

2hu12: @villander Could you please give me a example of how to test if something breaks? like running `npm run test`?

villander: @2hu12 can you finish this to be able to finish?

I would love see this merged

2hu12: @villander I just did some manual tests, the dummy app which has an engine called `ember-blog` with `lazyLoading: true` seems ok with the latest change and all tests pass locally.

stefanpenner:
@2hu12 and myself are kicking the tires.

tomdale:
@2hu12 Awesome work!

mydea:
Fyi, I've also put this into an example repo where the issue could be seen before - and it nicely fixes the duplication: https://github.com/mydea/test-ember-engines-addons 👍 Would love to see this land!

scalvert:
Super awesome work here, @2hu12! This has been a huge problem for us for a while. I can't wait to see this land.

stefanpenner:
@2hu12 I've rebased your PR.

lib/utils/deeply-non-duplicated-addon.js
```diff
@@ -0,0 +1,44 @@
...
+      return true;
+    }
+
+    if (addon.addons.length > 0) {
+      addon._orginalAddons = addon.addons;
+      deeplyNonDuplicatedAddon(hostAddons, addon, treeName);
```
stefanpenner: As implemented, it appears (although it is possible I have misread) we only prevent duplication when a `hostAddon` is included in one of the descendent addons. I believe this is actually incomplete, as duplicate descendent addons should also be handled.

Here is a more concrete example:

Given the following add-on dependency graph:

```
a -> b
a -> c
a -> d
b -> d 
d -> c
```

Assuming none of these are `hostAddons` I believe we will still duplicate `d` (via `a` and `b`) twice, and `c` (via `a`, `b`, `d`) three times.

lib/utils/recover-addons.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = function recoverAddons(addon) {
```
stefanpenner: nit: `restoreOriginalAddons`  may be a more clear method name

lib/engine-addon.js
```diff
@@ -330,36 +333,27 @@ module.exports = {
...
-              ) {
-                // the addon specifies cache key and it is the same as host instance of the addon, skip the tree
-                return;
-              }
-            } else {
-              return;
-            }
-          }
+      if (methodName === 'treeFor') {
+        recoverAddons(this);
```
stefanpenner: As we are doing some semi dangerous monkey-patching here, the restoration should occur within a `finally` clause.

lib/utils/deeply-non-duplicated-addon.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object} hostAddons
+ * @param {EmberAddon} dedupedAddon
+ * @param {String} treeName
+ */
+module.exports = function deeplyNonDuplicatedAddon(hostAddons, dedupedAddon, treeName) {
+  if (dedupedAddon.addons.length === 0) {
+    return;
+  }
+
+  dedupedAddon._orginalAddons = dedupedAddon.addons;
```
stefanpenner: Although this works in common situation, I believe it poses some broader risks. For example, if the patched `addon` instance is referred to in async or callback code, it will appear to have divergent `addons` arrays.

Although this scenario is rare, when it occurs it will be extremely surprising:

Contrived example to illustrate:

```js
treeForSomething(tree) {
  let addonsWhenHookCalled = this.addons;
  return new BroccoliPlugin(tree, {
    someCallbackInvokedLater: () => {
      let addons = this.addons; // imagine we need the list of addons here for some reason

      addons !== addonsWhenHookCalled; // this becomes surprising and incoherent
     }
  });
}
```

---

I've thought about how to fix this, I believe it can be done by utilizing a proxy, which wrap's an add-on, and traps when `addon.addons` is accessed, providing only the addons that should be visible. (And doing so recursively).

Does taking this approach sound reasonable?

lib/engine-addon.js
```diff
@@ -330,36 +333,27 @@ module.exports = {
       let invokeArguments = args || [];
       let hostAddons = this.ancestorHostAddons();
 
-      return this.addons
-        .map(addon => {
+      if (methodName === 'treeFor') {
+        deeplyNonDuplicatedAddon(hostAddons, this, invokeArguments[0]);
```
stefanpenner: Although in most cases this should be fine, I would like to recommend we put this new behavior behind a flag (off by default), and release as a minor version bump. After some time, we can swap the default to be true, and do a major version bump. Then at some future point in time, if all seems well, we can drop the flag entirely.

lib/utils/deeply-non-duplicated-addon.js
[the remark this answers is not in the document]
villander: here is another place that we can use `ensureLazyLoadingHash` util as well

lib/utils/deeply-non-duplicated-addon.js
villander: good gotcha @stefanpenner I had not realized that.

lib/utils/deeply-non-duplicated-addon.js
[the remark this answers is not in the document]
2hu12: I think the idea is if it encounters a child lazy engine inside a "deeplyNonDuplicatedAddon", it could just skip the resursive "deeplyNonDuplicatedAddon" for this child lazy engine for now and keep it, and when it goes to the child lazy engine's own "treeFor" method, it will pass through it's own "deeplyNonDuplicatedAddon".

For example, `father-lazy-engine-A` has `child-lazy-engine-B` and `normal-addon-C`, the "deeplyNonDuplicatedAddon" of `father-lazy-engine-A` will also recursively "deeplyNonDuplicatedAddon" `normal-addon-C` but `child-lazy-engine-B`, and then "child-lazy-engine-B" will have it's own "deeplyNonDuplicatedAddon".

I just unintentionally resovle this issue, please feel free to reopen it.

lib/utils/deeply-non-duplicated-addon.js
2hu12: For your example, if none of these are `hostAddons`, all of these will be `lazy engines addons`, so the situation is kinda like a normal ember app without engines.
Do you think the normal ember app will have the duplication in it's "vendor.js" if all of these addons are `app addons`? I think the anwser is no, I might find some realted code of "ember-cli", https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/merge-trees.js#L60 and https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L635 and https://github.com/ember-cli/ember-cli/blob/master/lib/models/addon.js#L654.

So the duplication this PR aim to address is between the host app and lazy engines and I think there's no duplication of addons issue within a normal app.

villander: @stefanpenner can you reproduce your scenario in a public repository? using https://github.com/kaliber5/ember-cli-bundle-analyzer

I think ember-engines should be concerned about this, since this use case is specific to the ecosystem of ember engines and isolation.

lib/engine-addon.js
2hu12: You mean the `Promise.prototype.finally` or some `finally` hook?

lib/utils/deeply-non-duplicated-addon.js
2hu12: Could any `treeForSomething` in `engine-addon.js` or ember-lic addon have async behaviour? It's seems everything is sync like `return addon[methodName].apply(addon, invokeArguments);` or it should return a Promise.
Is there any real world example rather than a contrived one?

lib/engine-addon.js
rwjblue:
```
try {
  // monkey patch here
} finally {
  // restore here
}
```

2hu12: Get it, thank you @rwjblue .

lib/utils/deeply-non-duplicated-addon.js
stefanpenner: @2hu12 contrived, and for us to discuss. If we don't feel it is worth it, we don't need to doit.

lib/utils/deeply-non-duplicated-addon.js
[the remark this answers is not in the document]
2hu12: It needs the comment indeed, it takes me sometime to recall the reason why.

lib/utils/deeply-non-duplicated-addon.js
stefanpenner: I'll take a look today.

lib/engine-addon.js
2hu12: If the flag is off, should it fall back to the old non deeply deduped implementation?

lib/utils/deeply-non-duplicated-addon.js
villander: @stefanpenner any new on this?

lib/engine-addon.js
rwjblue: Yes

lib/utils/deeply-non-duplicated-addon.js
stefanpenner: I believe @2hu12 raises a good point, this likely is not needed and if it is can be brought back in at a later time.

villander: Is not doubling addons a critical problem? I am convinced that yes, but we need to know where this should be solved. It should be resolved here or at ember-cli.

We can not miss the tracking of this, and we can open a issue.

rwjblue:
Looks like CI failures are due to ember-cli-addon-tests always using `emberjs/data#master` instead of a released version (the `emberjs/data` repo was recently changed into a monorepo so linking directly this way is invalid). Trying to fix that over in https://github.com/tomdale/ember-cli-addon-tests/pull/208...

lib/utils/deeply-non-duplicated-addon.js
2hu12: @villander I just had a verification on this, please check it out here https://github.com/2hu12/ember-engine-verification. It has a lazy engine `engine-a-lazy`, and three in-repo-addon `addon-a`, `addon-b` and `addon-c`, all three in-repo-addons are all not "hostAddon", and the dependency graph is like:
```
a -> c
b -> c
```
And the result is that it only has one piece of "addon-c" in https://github.com/2hu12/ember-engine-verification/blob/master/dist/engines-dist/engine-a-lazy/assets/engine-vendor.js#L29

stefanpenner: @2hu12 thanks for the verification!

stefanpenner:
@2hu12 2 more things, then we can merge:

* readme addition re: this flag
* CI green.

lib/utils/deeply-non-duplicated-addon.js
villander: @2hu12 thank you so much for explain this.

the duplicity of `hostAddons` should be a problem that the ember-cli team must deal with.

Many thanks for doing this here, ember-engines is much better now because of you.

2hu12:
> * [x]  readme addition re: this flag

rwjblue:
ember-cli-addon-tests@0.11.1 is released, and likely fixes the CI failures around ember-data packages missing

rwjblue:
https://github.com/ember-engines/ember-engines/pull/639 bumps the version independently, so we can confirm that CI is green on master...

villander:
@rwjblue @stefanpenner I think that we can merge this and generate `v0.8.0` release, because we know that CI error happens because `ember-cli-addon-tests`and don't impact anything on codebase that do ember-engines works ; )

gabrielcsapo:
@villander so excited for this! 

https://media.giphy.com/media/3o6UB3VhArvomJHtdK/giphy.gif

@2hu12 and everyone thanks for this!

villander:
@stefanpenner @rwjblue any news here?

scalvert:
If we can get #639 merged, we can get CI back to green. Right now that PR is failing independently. I'll try to poke at it to see what's going on.

villander:
thanks @scalvert

2hu12:
It seems that the CI is failed against ember-source@beta and canary for some incompatibility brought by [FEATURE ember-glimmer-angle-bracket-built-ins](https://github.com/emberjs/ember.js/pull/17772).

buschtoens:
@2hu12 Correct, `ember-beta` and `ember-canary` are failing because of the `<LinkTo>` PR. I have a fix here: #642.

However, there is still another regression in Ember: https://github.com/emberjs/ember.js/issues/17963


Don't know why `yarn test:node` is failing though.

2hu12:
@buschtoens Very glad you already find a way to fix this problem, I have some changes based on your version https://github.com/buschtoens/ember-engines/pull/23. Also, I have another PR https://github.com/emberjs/ember.js/pull/17971 which try to fix the remaining failed test.

scalvert:
What's currently blocking this PR? Anything I can help with?

villander:
@scalvert some guys are trying solve a issue com dynamic route properties on canary here #642 but without success.

2hu12:
@scalvert, @buschtoens has a PR for fixing the failing test of ember@beta and ember@canary, but the ember@beta also has a bug on `{{#link-to}}` component which might be fixed https://github.com/emberjs/ember.js/pull/17971, the fix still needs a test.

stefanpenner:
I am going to merge and release this, the test failures are unrelated to this PR and it provides good value as it is. The test failures are from sources outside of this repo, and are independently being moved forward.

I don't normally merge when red, but after careful consideration and review, and also that this has been open for so long, I am making an exception.

villander:
thanks @stefanpenner

stefanpenner:
released as v0.7.1 🎉thanks @2hu12  for both your patience, and pushing this forward.

buschtoens:
There's a related bug: #405

AFAICT it doesn't directly tie into this new `EMBER_ENGINES_ADDON_DEDUPE` pipeline, but it's closely related.

PR #596, opened 2018-10-03
POC: Don’t accidentally re-include the compact reexport plugin more then once

stefanpenner:
cc @rwjblue


stefanpenner:
I noticed this was causing us to mutate the share plugin state, resulting in babel-plugin-compact-reexports being included 1 per apps engine...

lib/engine-addon.js
```diff
@@ -423,9 +424,16 @@ module.exports = {
...
           this.options.babel.plugins = [];
+        } else {
+          // if we intend to mutate configuration provided to us, first we must
+          // clone it. Otherwise we may be mutating someone elses state.
+          this.options.babel.plugins = this.options.babel.plugins.slice();
+
         }
 
-        this.options.babel.plugins.push(require.resolve('babel-plugin-compact-reexports'));
+        if (this.options.babel.plugins.indexOf(BABEL_COMPACT_REEXPORTS_PLUGIN)) {
```
gabrielcsapo: Can we use includes?

stefanpenner: good call!

PR #597, opened 2018-10-03
Don’t accidentally re-include the compact reexport plugin more then once

stefanpenner:


stefanpenner:
released as v0.5.24 🎉

PR #600, opened 2018-10-24
fix(router-ext): hook into `getRoute`, if exsits

### LINKED ISSUE #599: ember-source@v3.6.0: Router#getHandler is deprecated and not callled any more
[`Router#getHandler`](https://github.com/emberjs/ember.js/blob/v3.6.0-beta.1/packages/@ember/-internals/routing/lib/system/router.ts#L216-L230) has been [deprecated](https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_remove-handler-infos) in `v3.6.0-beta.1`. It was still available in [`v3.5.0`](https://github.com/emberjs/ember.js/blob/v3.5.0/packages/ember-routing/lib/system/router.js#L565-L611).

ember-engines hooks into it here:

https://github.com/ember-engines/ember-engines/blob/d44802d05054919b3f701a00911267d597cc27e9/addon/-private/router-ext.js#L60-L69

Apparently, `getHandler` is not being called any more in `v3.6.0-beta.1`, which means that the hook is never executed. As a consequence, lazy-loaded engines aren't loaded and users see this unhelpful error:

```
Uncaught TypeError: Cannot read property 'lookup' of undefined
    at PrivateRouter.getRoute (router.js:180)
    at UnresolvedRouteInfoByParam.fetchRoute (router_js.js:617)
    at UnresolvedRouteInfoByParam.get (router_js.js:648)
    at URLTransitionIntent.applyToState (router_js.js:1130)
    at PrivateRouter.getTransitionByIntent (router_js.js:1267)
    at PrivateRouter.transitionByIntent (router_js.js:1224)
    at PrivateRouter.doTransition (router_js.js:1335)
    at PrivateRouter.handleURL (router_js.js:1643)
    at Class._doURLTransition (router.js:460)
    at Class.handleURL (router.js:456)
```

This is because `engineInstance` is `undefined` in [`PrivateRouter#getRoute`](https://github.com/emberjs/ember.js/blob/v3.6.0-beta.1/packages/@ember/-internals/routing/lib/system/router.ts#L290-L295), since [`RouterExt#_loadEngineInstance`](https://github.com/ember-engines/ember-engines/blob/d44802d05054919b3f701a00911267d597cc27e9/addon/-private/router-ext.js#L86-L93) was never called.

buschtoens:
Fixes #599.

```js
 // This method used to be called `getHandler` and it is going to be called `getRoute`.
```

The comment says it all. If only I was able to read comments. 🙏 😂

### DIFF
--- a/addon/-private/router-ext.js
+++ b/addon/-private/router-ext.js
@@ -61,7 +61,9 @@ Router.reopen({
     let isSetup = this._super(...arguments);
     if (newSetup) {
       // This method used to be called `getHandler` and it is going to be called `getRoute`.
-      if (this._routerMicrolib.getHandler !== undefined) {
+      if (this._routerMicrolib.getRoute !== undefined) {
+        this._routerMicrolib.getRoute = this._handlerResolver();
+      } else if (this._routerMicrolib.getHandler !== undefined) {
         this._routerMicrolib.getHandler = this._handlerResolver();
       }
     }


buschtoens:
I've tested this branch with our application and can confirm that the fix works. 😊 👍

PR #611, opened 2019-01-13
bump babel-plugin-compact-reexports

villander:


villander:
It's was updated fixing support to node 4 and created a new release. How it's was created on-demand to ember-engines, I found it interesting to keep updated.

https://github.com/ember-engines/babel-plugin-compact-reexports/pull/10/files

PR #613, opened 2019-01-14
Make lazy routable acceptance test more forgiving

dgeb:
Makes a lazy-loading acceptance test more forgiving of the number of bundle load errors that are encountered on different versions of ember. Three are encountered in 3.7, while only two are encountered in 3.3. This difference does not seem to impact the behavior of the addon, yet does prevent CI from passing.


rwjblue:
Awesome, thank you!

PR #616, opened 2019-01-15
Bump broccoli-dependency-funnel

villander:
Good improvement - https://github.com/ember-engines/broccoli-dependency-funnel/pull/32/files


dgeb:
Thanks @villander!

PR #631, opened 2019-03-26
Avoid errors when keywords array is not present.

rwjblue:
The recent addition of the keyword check throws an error (`Cannot read property 'indexOf' of undefined`) if keywords are not present. This fix ensures that the warning is issued either way (with or without a keywords array) and also adds tests that would have previously failed.


stefanpenner:
:shipit:

PR #632, opened 2019-03-27
Ensure `postprocessTree` to ran before `broccoli-asset-rev`

villander:
Fixed the true scenario - https://github.com/ember-engines/ember-engines/pull/384#issuecomment-321823354

@rwjblue @dgeb

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "node": "6.* || 8.* || >= 10.*"
   },
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "before": "broccoli-asset-rev"
   }
 }


villander:
@dgeb can we merge this?

villander:
@stefanpenner can you take a look this, please?

villander:
@stefanpenner @rwjblue  can you take a look this, please?

dgeb:
This looks identical to #384 🤔

villander:
no @dgeb I replace `after` per `before` following the true scenario by @rwjblue on old PR.

take a look, please - https://github.com/ember-engines/ember-engines/pull/384#issuecomment-321823354

dgeb:
> no @dgeb I replaced after per before following the true scenario by @rwjblue on old PR.

Yeah, I finally figured it out after staring at the diff again (and deleted my prior comment). Thanks for the clarification.

buschtoens:
Did you actually try running this with a production build for lazy engines?

This is the content of the `<meta name="[app-name]/config/asset-manifest">` tag, when I use the latest `master` of `ember-engines`.

```js
{
  "bundles": {
    "@clark/engine-cockpit-retirement": {
      "assets": [
        {
          "uri": "/engines-dist/@clark/engine-cockpit-retirement/assets/engine-vendor.js",
          "type": "js"
        },
        {
          "uri": "/engines-dist/@clark/engine-cockpit-retirement/assets/engine.css",
          "type": "css"
        },
        {
          "uri": "/engines-dist/@clark/engine-cockpit-retirement/assets/engine.js",
          "type": "js"
        }
      ]
    },
    // ...
  }
}
```

As you can see, the URIs are not fingerprinted in there and loading the assets fails as the actual files _are_ fingerprinted.

buschtoens:
Interestingly the correct "fix" for me actually is #384. It only works with `after`, not `before`.

buschtoens:
> Even more interestingly, just removing the `"before": "broccoli-asset-rev"` from the latest `master` suddenly exhibits the same issue, even though this setup has been working for months.

The reason for this was that I used a pre-published version as `@buschtoens/ember-engines`, which lexicographically sorts it before `broccoli-asset-rev`. When using something like [`patch-package`](https://github.com/ds300/patch-package#readme) instead, this does not occur.

tomwayson:
I agree w/ @buschtoens, I think the correct way would be to use `"after": "broccoli-asset-rev"`.

First, that's [what ember-asset-loader says is required](https://github.com/ember-engines/ember-asset-loader/blob/v0.6.1/README.md#using-with-broccoli-asset-rev).

Secondly, we [started having issues lazy-loading fingerprinted engine assets after upgrading our app to ember-auto-import 2.x](https://github.com/embroider-build/ember-auto-import/issues/609). The filenames in the dist folder _were_ fingerprinted, but the entries in the asset manifest (both in index.html and asset-manifest.json) were **not**.

After debugging our build and reading through issues in all the related addon repos, it really seemed to me that _somehow_ the dependency bump changed the order in which the addons were run. To verify, I added `"after": "broccoli-asset-rev"` to ember-engine's package.json (locally w/in our node_modules folder) and now the asset manifest URIs are fingerprinted.

My guess is that the current situation (w/o specifying `before` or `after`) works by happenstance unless your dependencies change the lexicographical sorting  of the addons and the order in which they are run.

I think https://github.com/ember-engines/ember-engines/pull/384#pullrequestreview-55812394 is also correct, when using `after` we do not get the other features of broccoli-asset-rev, like `prepend`ing. That has been true for us even before it recently broke, but we address that by using it's `generateURI()` hook.

In order to use `before` work the way that @rwjblue describes above, I think you'd have to also add  .json to broccoli `replaceExtensions`, which may have unintended consequences.

PR #633, opened 2019-03-27
Copy of Load config based on `storeConfigInMeta` option

puopg:
Related: #228 

This is an updated version of #228, tests passing

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -982,7 +982,7 @@ module.exports = {
       @param type
     */
     options.contentFor = function(type, config) {
-      if (type === 'head') {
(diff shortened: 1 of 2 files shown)


lib/engine-addon.js
```diff
@@ -1033,9 +1033,17 @@ module.exports = {
     options.getEngineConfigContents =
       options.getEngineConfigContents ||
       function() {
-        const configTemplate = fs.readFileSync(`${__dirname}/engine-config-from-meta.js`, 'UTF8');
+        var storeConfigInMeta = this.app.options.storeConfigInMeta;
```
villander: We need of tests here

PR #638, opened 2019-04-10
Update ember-cli-addon-tests to use ember-datav3.8.0

kiwiupover:
https://github.com/tomdale/ember-cli-addon-tests/pull/208

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "ember-blog": "link:./tests/dummy/lib/ember-blog",
     "ember-chat": "link:./tests/dummy/lib/ember-chat",
     "ember-cli": "~3.7.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "ember-blog": "link:./tests/dummy/lib/ember-blog",
     "ember-chat": "link:./tests/dummy/lib/ember-chat",
     "ember-cli": "~3.7.1",
-    "ember-cli-addon-tests": "^0.11.0",
+    "ember-cli-addon-tests": "github:kiwiupover/ember-cli-addon-tests#test-against-ember-data-release",
```
villander: What is the meaning of this PR? What does it solve?

we should use stable releases here

kiwiupover: closed by #639

Issue #540, opened 2018-04-11
Documentation


MelSumner:
Expected: 
The documentation would be consolidated to one place and when new features are added/updated, they are added to the documentation. 

Actual:
Right now, using ember-engines requires a bit of juggling between Github and the guides, which can give the impression that documentation is missing. 

Proposed solution: 

1. Make the README.md a table of contents and link to sections on the ember-engines guides/site. 
2. When something new is merged, it can be added to the README (table of contents), even if the author doesn't have time to write the documentation. This will also serve to identify where documentation still needs to be added. 
3. Add a roadmap section to make it visible what the "roadmap to 1.0" looks like

villander:
@MelSumner can we keep ember-engines site or will to do a new site with [Ember Addons Docs](https://github.com/ember-learn/ember-cli-addon-docs)?

mydea:
I would like to maybe contribute some things to the engines docs, but am unsure about where to go/how to start.

MelSumner:
@villander check out https://github.com/ember-engines/ember-engines.com that already exists.

PR #641, opened 2019-04-15
Replace old docs with table of contents

villander:
It's solves the #540  🎉 

cc: @MelSumner @dgeb @rwjblue @stefanpenner


villander:
@rwjblue this too, please  ; )

PR #642, opened 2019-04-23
fix(link-to): support angle-bracket invocation

buschtoens:
[RFC #0459](https://emberjs.github.io/rfcs/0459-angle-bracket-built-in-components.html) changed the attribute name from `targetRouteName` to `route`.

Here's the PR: https://github.com/emberjs/ember.js/pull/17772

This is currently breaking the Ember v3.10.0-beta series.

PR #647, opened 2019-05-17
[fix] prefix route name to `route` in <LinkTo>

lennyburdette:
In ember-source 3.10, support for "angle bracket built-in" components like <LinkTo> was enabled. In the refactoring, the `targetRouteName` property was renamed to `_route `. This change overrides the `_route ` property to add the engine prefix (in addition to the original `targetRouteName` prefix, so that we continue to support Ember 3.9 and below.)

targetRouteName for curly {{link-to}} components:
https://github.com/emberjs/ember.js/blob/eee02fb72a47ff1ffbc9919fcf3fc4b74b937093/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L1765

_route for angle-bracket <LinkTo> components:
https://github.com/emberjs/ember.js/blob/eee02fb72a47ff1ffbc9919fcf3fc4b74b937093/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L490

### DIFF
--- a/addon/components/link-to-component.js
+++ b/addon/components/link-to-component.js
@@ -16,8 +16,13 @@ export default LinkComponent.extend({
     );
 
     if (owner.mountPoint) {
-      // Prepend engine mount point to targetRouteName
+      // [Ember <= 3.9] Prepend engine mount point to targetRouteName
+      // (for {{link-to "..."}})
       this._prefixProperty(owner.mountPoint, 'targetRouteName');
+
+      // [Ember >= 3.10] Prepend engine mount point to route
+      // (for <LinkTo @route="...">)
+      this._prefixProperty(owner.mountPoint, 'route');
 
       // Prepend engine mount point to current-when if set
       if (get(this, 'current-when') !== null) {
--- a/addon/components/link-to-external-component.js
+++ b/addon/components/link-to-external-component.js
@@ -9,8 +9,10 @@ export default LinkComponent.extend({
     const owner = getOwner(this);
 
     if (owner.mountPoint) {
-      const targetRouteName = get(this, 'targetRouteName');
+      // `route` for Ember >= 3.10, `targetRouteName` for Ember <= 3.9
+      const targetRouteName =  get(this, 'route') || get(this, 'targetRouteName');
       const externalRoute = owner._getExternalRoute(targetRouteName);
+      set(this, 'route', externalRoute);
       set(this, 'targetRouteName', externalRoute);
     }
   },


lennyburdette:
This change fixed my reproduction: https://github.com/lennyburdette/engines-310/tree/fixed

lennyburdette:
This solution triggers a deprecation warning:
```
DEPRECATION: The <eager-blog@component:link-to::ember207>#_route computed property was just overriden. This removes the computed property and replaces it with a plain value, and has been deprecated. If you want this behavior, consider defining a setter which does it manually. [deprecation id: computed-property.override] See https://emberjs.com/deprecations/v3.x#toc_computed-property-override for more details.
```

lennyburdette:
This branch fixes the bug, but tests don't pass because of an extremely subtle behavior change with `<LinkTo>` components:

* The original `{{link-to}}` `qualifiedRouteName` property has a subtle "bug": it uses the `_routing.currentRouteName` as a fallback value if `targetRouteName` is missing, but it **doesn't** [list it as a dependency key](https://github.com/emberjs/ember.js/blob/eee02fb72a47ff1ffbc9919fcf3fc4b74b937093/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L1618-L1620). 
   
    This is important only for links that rely on the current route, like `{{link-to (query-params lang="Japanese")}}`. The effect is that the `qualifiedRouteName` property doesn't change once set, even if the app routes to a new location and `_routing.currentRouteName` changes.
* The new `<LinkTo>` `_route` property "correctly" [lists `_routing.currentRouteName` as a dependency key](https://github.com/emberjs/ember.js/blob/eee02fb72a47ff1ffbc9919fcf3fc4b74b937093/packages/%40ember/-internals/glimmer/lib/components/link-to.ts#L490-L493), so it recalculates the link's `href` when the app routes to a new location.

   In the [failing test](https://github.com/ember-engines/ember-engines/blob/master/tests/acceptance/routeable-engine-demo-test.js#L490-L509), the app routes to an error subroute, so the `{{link-to (query-params lang="Japanese")}}` tries recalculates its `href` to reflect the error route (which doesn't actually change the url.) 

    This shouldn't be an issue, but the fabricated error substate gets the url suffix: [`"/_unused_dummy_error_path_route_" + name + "/:error"`](https://github.com/emberjs/ember.js/blob/83263728b116fac6d4e8fcc798babf555a62ac2b/packages/%40ember/-internals/routing/lib/system/dsl.ts#L76). The addition of the `:error` path part makes the URL recognizer fail because `{{link-to (query-params lang="Japanese")}}` doesn't provide a value for `:error`. The router throws an error, which fails the test.

I've got no idea what to do ... wdyt @rwjblue @chancancode ?

villander:
@dgeb do you have any idea of how to do it? We just need fix this [failing test](https://github.com/ember-engines/ember-engines/blob/master/tests/acceptance/routeable-engine-demo-test.js#L490-L509)

2hu12:
I think we already have this one https://github.com/ember-engines/ember-engines/pull/642 for the same issue.

rwjblue:
@lennyburdette - Thanks for working on this (you too @2hu12!), I just merged https://github.com/ember-engines/ember-engines/pull/642 which was opened a bit before this one and should fix the underlying issue.

Again, sorry for the delays here.

Issue #98, opened 2016-05-03
Allow arguments to `{{mount}}` template keyword and `mount` routing helper


dgeb:
The following was discussed at the Ember Core Team F2F in January:

> The {{mount}} template keyword should allow normal handlebars style hash arguments. The router DSL style should also allow some sort of API to supply parameters too.

Still TBD:
- What API should the `mount` routing helper support for accepting parameters?
- Are arguments set directly on the engine instance?
- How can an engine declare which properties are writeable?

trentmwillis:
Out of curiosity, what are some use cases this is attempting to solve? My perspective is that passing in values like this would be akin to passing state which I assumed is what we had service dependencies for.

mike183:
@trentmwillis Our use case for this would be for the dashboard of the CMS we're developing.

When the dashboard loaded we would retrieve a list of widgets that needed to be displayed, we would then loop through each of the widgets mounting them onto the dashboard. When mounting the widget, we would also pass through some initial config/data for the widget to use.

The data passed through could be some initial data to display, preventing each widget needing to make a separate request to the server for data.

While simplified for the sake of brevity, we would have something like this:

```hbs
{{#each dashboardWidgets as |widget|}}
  {{mount widget.name config=widget.config data=widget.data}}
{{/each}}
```

trentmwillis:
@mike183 is there a reason why a service couldn't mediate that transfer of information?

dgeb:
@trentmwillis Usually the need for this comes down to a single engine being used in multiple contexts within the same app. Passing a parameter to the engine could configure its context.

Take for example a routable "blog" engine which is used to create a "design" blog mounted at `/design` and a "developer" blog mounted at `/dev`. Even if a common service provides data for both blogs, each blog instance will need to be passed some configuration parameters to properly filter data from that service.

mike183:
@trentmwillis The dashboard widgets would be developed by third parties so it would just be a case simplifying the development of the widgets and reducing boilerplate code.

@dgeb That's a pretty interesting use case, I'd never considered using engines in such a way.

trentmwillis:
> a single engine being used in multiple contexts within the same app.

Okay, that use case makes sense to me, since it doesn't make sense to have the branching logic for multiple configs within the Engine.

chadhietala:
My concern is that you could introduce runtime dependencies making tree shaking and lazy loading more difficult.

```javascript
export default Ember.Component.extend({
[...]

PR #650, opened 2019-05-31
Update README.md

garjust:
Correct `{{mount}}` documentation. A model can be now be passed and is documented in Ember documentation for the mount helper.

Related open issue: https://github.com/ember-engines/ember-engines/issues/98


rwjblue:
Thank you!

Issue #525, opened 2018-03-13
Support for new test syntax


mydea:
Has anyone got integration tests working for their (lazy) engine, with the new test syntax?

I get the error that `xxxx is not a helper`. I guess it is because `setApplication` sets the dummy app, in which the components themselves are not directly registered?

I somehow got it to work with this pretty hacky code:

```js
import Resolver from '../../resolver';
import { setResolver } from '@ember/test-helpers';

const modulePrefix = 'my-engine-name';
const resolver = Resolver.create();

export default function setupResolver(hooks) {
  // Manually fix the entries for the engines so they can be found by the resolver
  // Put everything from the my-engine-name namespace into dummy as well
  // E.g. my-engine-name/components/my-component becomes dummy/components/my-component
  let modules = resolver._moduleRegistry._entries;

  let newModules = {};
  let overwriteModules = {};
  let originalModules = {};

  Object.keys(modules).forEach((module) => {
    if (module.startsWith(modulePrefix)) {
      let newName = `dummy/${module.substr(modulePrefix.length + 1)}`;

      if (!modules[newName]) {
        newModules[newName] = modules[module];
      } else {
        originalModules[newName] = modules[module];
        overwriteModules[newName] = modules[module];
      }
    }
  });

  hooks.beforeEach(function() {
    Object.keys(newModules).forEach((module) => {
      modules[module] = newModules[module];
    });
    Object.keys(overwriteModules).forEach((module) => {
      modules[module] = overwriteModules[module];
    });
  });

  hooks.afterEach(function() {
    Object.keys(newModules).forEach((module) => {
      delete modules[module];
    });
    Object.keys(overwriteModules).forEach((module) => {
      modules[module] = originalModules[module];
    });
  });

  return setResolver(resolver);
}
```

and then, in my integration tests:

```js
import { module, test } from 'qunit';
import { setupRenderingTest } from 'ember-qunit';
import setResolver from 'dummy/tests/helpers/set-resolver';
// ... other dependencies

module('Integration | Component | my test component', function(hooks) {
  setupRenderingTest(hooks);
  setResolver(hooks);

  // actual tests go here
});
```

I'm sure there is a better way, but I'm not sure what it is ;)

mydea:
I also seem to be unable to overwrite the store service in integration tests in the engine. E.g. this doesn't work, the store service is always the 'real' store service:

```js
const storeService = Service.extend({
  // my mock service
});

module('Integration | Component | my-component', function(hooks) {
[...]

PR #653, opened 2019-06-09
Improve Test API for Engines

villander:
This improves Test API for ember-engines, we need to document this on Engine Guides 🎉 

Introduces two new APIs:

* `loadEngine` -- Loads the specified engine (including assets), instantiates it, and returns the Engine's "owner" via `this.engine`.
* `setupEngine` -- A convenient wrapper around `loadEngine` that works nicely in the same style as `ember-qunit`.

---

#### Rendering tests
```js
import { module, test } from 'qunit';
import { setupRenderingTest } from 'ember-qunit';
import { render, click } from '@ember/test-helpers';
import hbs from 'htmlbars-inline-precompile';
import { setupEngine } from 'ember-engines/test-support';

module('Integration | Component | setupEngineRenderingTest', function (hooks) {
  setupRenderingTest(hooks);
  setupEngine(hooks, 'ember-blog');

  test('should change colors', async function(assert) {
    assert.expect(2);

    // set the outer context to red
    this.set('colorValue', 'red');

    await render(hbs`<PrettyColor @name={{this.colorValue}} />`, { owner: this.engine });

    assert.equal(this.element.querySelector('[data-test-pretty-color]').textContent.trim(), 'Pretty Color: red', 'starts as red');

    this.set('colorValue', 'blue');

    assert.equal(this.element.querySelector('[data-test-pretty-color]').textContent.trim(), 'Pretty Color: blue', 'updates to blue');
  });

  test('should update title on button click', async function(assert) {
    assert.expect(2);

    await render(hbs`<MagicTitle />`, { owner: this.engine });

    assert.equal(this.element.querySelector('h2').textContent.trim(), 'Hello World', 'initial text is hello world');

    // Click on the button
    await click('[data-test-title-button]');

    assert.equal(this.element.querySelector('h2').textContent.trim(), 'This is Magic', 'title changes after click');
  });
});
```

#### Unit tests 

```js
import { module, test } from 'qunit';
import { setupTest } from 'ember-qunit';
import { setupEngine } from 'ember-engines/test-support';

module('Unit | Route | post.likes', function (hooks) {
  setupTest(hooks);
  setupEngine(hooks, 'ember-blog');

  test('it exists', function (assert) {
    assert.expect(1);

    let route = this.engine.lookup('route:post.likes');
    assert.ok(route);
  });
});
```

#### Application tests

```js
import { module, test } from 'qunit';
import { visit, currentURL } from '@ember/test-helpers';
import hbs from 'htmlbars-inline-precompile';
import { setupApplicationTest } from 'ember-qunit';
import { setupEngine } from 'ember-engines/test-support';

module('basic full application test', function(hooks) {
  setupApplicationTest(hooks);
  // no setupEngine in application tests, because they should load the engine automatically

  test('the user can visit blog engine', async function(assert) {
    await visit('/routable-engine-demo/blog/new');

    assert.equal(currentURL(), '/routable-engine-demo/blog/new');
  });
});

### CODE REVIEW

**Comment by villander** [comment]
@rwjblue I did your solution, but it don't work

```js
export function setupEngineTest(hooks, engineName) {
  hooks.beforeEach(function() {
    if (!this.owner) {
      throw new Error('You must call one of the ember-qunit setupTest(),'
        + ' setupRenderingTest() or setupApplicationTest() methods before'
        + ' calling setupEngineTest()');
    }

    let resolver = engineResolverFor(engineName)

    setResolver(resolver);
  });
}
```

I get this error:

image

rwjblue:
FYI - @villander and I are working on this together, we have a nice working solution for `setupRenderingTest` and `setupTest`. Working on making `setupApplicationTest` a bit nicer still....

villander:
#### Integrations tests
```js
import { module, test } from 'qunit';
import { render } from '@ember/test-helpers';
import hbs from 'htmlbars-inline-precompile';
import { setupRenderingTest } from 'ember-qunit';
import { setupEngineTest } from 'ember-engines/test-support';

module('Integration | Component | hello-name', function(hooks) {
  setupRenderingTest(hooks);
  setupEngineTest(hooks, 'ember-blog');

  test('it renders', async function(assert) {
    // Set any properties with this.set('myProperty', 'value');
    // Handle any actions with this.set('myAction', function(val) { ... });

    await render(hbs`{{hello-name name='Michael'}}`);

    assert.equal(this.element.textContent.trim(), 'Hello Michael');

    // Template block usage:
    await render(hbs`
      {{#hello-name}}

      {{/hello-name}}
    `);

    assert.equal(this.element.textContent.trim(), 'Hello');
  });
});
```

#### Unit tests 

```js
import { module, test } from 'qunit';
import { setupTest } from 'ember-qunit';
import { setupEngineTest } from 'ember-engines/test-support';

module('Unit | Route | index', function(hooks) {
  setupTest(hooks);
  setupEngineTest(hooks, 'ember-blog');

  test('it exists', function(assert) {
    let route = this.owner.lookup('route:index');
    assert.ok(route);
  });
});
```

addon-test-support/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Used to set up engine resolver for a unit test.
```
rwjblue: These docs are a bit wrong now (given the new implementation)

rwjblue:
Also, we need to finish off https://github.com/emberjs/ember-test-helpers/pull/664 before this will actually work properly.

addon-test-support/index.js
villander: fixed!

villander:
## for eager engines

```js
module("Asdf", function(hooks) {
  setupRenderingTest(hooks);
  setupEngineTest(hooks, 'engine-name-here');
})
```

## for lazy engines

```js
module("Asdf", function(hooks) {
  setupRenderingTest(hooks);
  setupEngineTest(hooks,  'engine-name-here', 'mount-point');
})
```

villander:
@rwjblue this solution it's good? if yes, I can finish `emberjs/ember-test-helpers#664` to merge this as well

rwjblue:
I don't think we should need the mount point, I'd like to figure out why that was necessary and remove it before we land this.

villander:
@rwjblue any update here? do you have any idea to load lazy engine without `mountPoint`?

villander:
@rwjblue any news?

villander:
> I don't think we should need the mount point, I'd like to figure out why that was necessary and remove it before we land this.

@rwjblue can you figure out if is possible finish this without `mount point`, please? I'd like of turn test API for engines more friendly

step2yeung:
@villander Thanks for pushing this forward!
Few questions:
1. I see we are using beforeEach to load the engine, am I correct in assuming this is so on each test, we will setup a clean instance of the engine?
2. I do not see any unload of the engine on an `afterEach` or `after` hook. So as we run many engine tests, more engines will be loaded and the environment will be polluted with engines loaded from previous tests?? I'm thinking if we should unload on `after` hook 
3. Since I am not involved earlier, how does https://github.com/emberjs/ember-test-helpers/pull/664 play along with this? Do we need to land that PR first?

rwjblue:
> How does https://github.com/emberjs/ember-test-helpers/pull/664 play along with this? Do we need to land that PR first?

Yes, we need to land that first. Specifically, without that change it is not possible to render components as if you were within the Engine instance (and therefore the components would not be found by the normal resolution process).

brendenpalmer:
Just curious, but what's the status on this PR? It seems like we were waiting on https://github.com/emberjs/ember-test-helpers/pull/664 to land?

SergeAstapov:
@villander tried this approach in our app and had to remove `preloadAssets(manifest).then(start)` from `tests/test-helper.js` for lazy loading engine.

Otherwise I get error
```
Error: Assertion Failed: invalid engine: missing mountPoint
```

@villander do you have any idea what I'm missing?

villander:
@SergeAstapov are you using this custom branch for ember-test-helpers https://github.com/emberjs/ember-test-helpers/pull/664?

SergeAstapov:
@villander I've used latest release of @ember/test-helpers with fixes from emberjs/ember-test-helpers#664 applied.

SergeAstapov:
@villander BTW this is only an issue for component rendering test, controller unit test works just fine

villander:
this must work for integration and acceptance tests as well.

SergeAstapov:
@villander the difference is in `this.owner.hasRegistration(`engine:${engineName}`)`:
* when assets are pre-loaded via `preloadAssets(manifest).then(start)` it evaluates to `true`
* when I have `start()` and assets not pre-loaded, it evaluates to `false` hence invokes `loadEngine()`. 

I assume difference comes from the fact that in first case it invokes `this.owner.buildChildEngineInstance(engineName)` and in the second - it executes `router._loadEngineInstance()`

I'm happy to discuss details in Discord if that could help.

TheLooseCannon:
@villander im finding it hard to setup my tests for the engines... how far along is your standalone helpers? is it possible for me to pull your branch and use it? or if this will be merged with the master soon, should i wait till that happens?

villander:
@TheLooseCannon you can pull my branch and reuse it, but don't forget of the `ember-test-helpers`. emberjs/ember-test-helpers#664?

villander:
@SergeAstapov I had to replace `preloadAssets(manifest).then(start)` with `start()` in tests/test-helper.js for lazy loading engine.

And it continues working fine, check it out the tests.

image

SergeAstapov:
@villander yeah, that seem to make the trick. If that's expected and acceptable, then no issue here.
Just an idea - this need to be referenced in docs somehow, I assume we could do so once this PR lands.

monovertex:
Following a short discussion with @villander on Discord, I wanted to raise a concern regarding this update: during acceptance tests I imagine that multiple engines might be necessary, in which case setting engine instances on `this.engine` can potentially cause some issues when access to those engines' contexts is necessary.

I propose using a registry of sorts instead, so that  `setupEngineTest` could be executed multiple times. Thoughts?

rwjblue:
> during acceptance tests I imagine that multiple engines might be necessary

Why would you need to access the engine instance in an acceptance test? In general, acceptance tests _already_ work without modification as the engine content is loaded just like it would be loaded in a normal production application (either `this.mount` or `{{mount`).

monovertex:
> > during acceptance tests I imagine that multiple engines might be necessary
> 
> Why would you need to access the engine instance in an acceptance test? In general, acceptance tests _already_ work without modification as the engine content is loaded just like it would be loaded in a normal production application (either `this.mount` or `{{mount`).

My specific case is that I have a service that uses `window.location` to redirect to an external URL, in an engine. In an acceptance test in the host app I need to stub this behavior, because `window.location` is read-only and currently I have no clean way of doing this. 

I think there could be other edge cases where stubbing services inside an engine is still needed, so if it's not hard to give a test access to an arbitrary engine's context, why not give developers the freedom to do it?

Finally, from what I understand from the comments on this PR, you still need to use `setupEngineTest` in acceptance tests, which will set up `this.engine`. So this becomes a pitfall for a developer that might rely on it, then they add a second `setupEngineTest` for engine B for some reason, and suddenly the test now fails.

rwjblue:
> you still need to use setupEngineTest in acceptance tests, at least with lazy engines, which will set up this.engine

No, you shouldn't need `setupEngineTest` at all in acceptance tests.

> My specific case is that I have a service that uses window.location to redirect to an external URL, in an engine. In an acceptance test in the host app I need to stub this service, because window.location is read-only and currently I have no clean way of doing this.

I would expect that you either have a `window` service in the app that is provided to each of your engines (and that would naturally flow down into the engines without any custom munging in the engine itself) or you would use something like [ember-window-mock](https://github.com/kaliber5/ember-window-mock) to avoid needing any specialized code at all.

> The idea of having a single this.engine is flawed if setupEngineTest can be called multiple times.

Maybe? Not sold personally, but we can certainly prevent calling it multiple times and avoid the foot gun.

villander:
> In general, acceptance tests already work without modification as the engine content is loaded just like it would be loaded in a normal production application

I totally agree!

Maybe we need to load two engines to test flow in acceptance tests? I'm not sure, will depend on how you structure your architecture

buschtoens:
btw we are currently using this test helper:

```ts
import { getApplication } from '@ember/test-helpers';
import { setupRenderingTest as _setupRenderingTest } from 'ember-qunit';

import { set } from '@ember/object';

import LinkToExternalComponent from 'ember-engines/components/link-to-external-component';import engineResolverFor from 'ember-engines/test-support/engine-resolver-for';

import TestContext from '../test-context';

const modulePrefix = 'some-engine';
const resolver = engineResolverFor(modulePrefix);

// https://github.com/ember-engines/ember-engines/issues/525#issuecomment-402131973
export default function setupRenderingTest(hooks: NestedHooks) {
  _setupRenderingTest(hooks, { resolver });
  const app = getApplication();

  hooks.beforeEach(function (this: TestContext) {
    this.owner.__registry__.fallback.fallback = app.__registry__;
    const config = this.owner.resolveRegistration('config:environment');
    if (config) {
      set(config, 'environment', 'test');
    }

    this.owner.register('component:link-to-external', LinkToExternalComponent);
  });
}
```

The interesting addition compared to https://github.com/ember-engines/ember-engines/issues/525#issuecomment-402131973 is 

```js
const app = getApplication();
this.owner.__registry__.fallback.fallback = app.__registry__;
```

This make render tests work with service injections of addons like `ember-responsive` or `ember-intl`, that define their own custom registrations, like `breakpoints:main` and `formats:main` respectively.

ShaharAvigezer:
jumping in after we encountered the issue ourselves on our team.
the file @buschtoens sent, alongside the replies here, really helped and saved us tons of hours.
Here is our version with some refactoring:

```javascript
import { getApplication } from '@ember/test-helpers';
import { setupRenderingTest } from 'ember-qunit';
import { set } from '@ember/object';
import engineResolverFor from 'frontend/tests/engine-resolver-for';

// https://github.com/ember-engines/ember-engines/issues/525#issuecomment-402131973
export function setupEngineRenderingTest(hooks, modulePrefix) {
  const resolver = engineResolverFor(modulePrefix);
  setupRenderingTest(hooks, { resolver });
  const app = getApplication();

  hooks.beforeEach(function () {
    this.owner.__registry__.fallback.fallback = app.__registry__;
    const config = this.owner.resolveRegistration('config:environment');
    if (config) {
      set(config, 'environment', 'test');
    }
  });
}
```

notice we added the engine module prefix to support multiple engines.
I hope this will be a part of the ember engine testing API soon

amk221:
Does this PR fix this?

```javascript
this.owner.setupRouter()  //  this.owner.setupRouter is not a function
```

lougreenwood:
Is there any progress on this? AFAIU, it removes the need for overriding the resolver, which causes issues with mirage in integration tests.

Are we just waiting on https://github.com/emberjs/ember-test-helpers/pull/795 to be completed now?

rwjblue:
@amk221 

> Does this PR fix this?
>
> ```js
> this.owner.setupRouter()  //  this.owner.setupRouter is not a function
> ```

Yes.

rwjblue:
We discussed this at todays ember-engines team meeting, and are trying to "reboot" this and get it landed soon.

villander:
After our monthly meeting today, we decide to try this possible API to make our tests on Engines simpler, and more likely the usual tests are already written to a normal Ember app or addon. Also once it landed the [preloadAssets](https://github.com/ember-engines/ember-asset-loader#pre-loading-assets-during-testing) ( _which is unrealistic, give the real flow the engines are loaded on-demand_ ) will be needless 

```js
module('setupEngineTest', function (hooks) {
    setupTest(hooks);
    hooks.beforeEach(async function() {
      this.engine = await loadEngine('my-engine-name');
    });
    test('it exists', function (assert) {
      assert.expect(1);

      let route = this.engine.lookup('route:post.likes');
      assert.ok(route);
    });
  });

  module('setupEngineRenderingTest', function (hooks) {
    setupRenderingTest(hooks);

    hooks.beforeEach(async function() {
      this.engine = await loadEngine('my-engine-name');
    });

    test('should change colors', async function(assert) {
      assert.expect(2);

      // set the outer context to red
      this.set('colorValue', 'red');

      await render(hbs`{{pretty-color name=colorValue}}`, { owner: this.engine });

      assert.equal(this.element.querySelector('div').getAttribute('style'), 'color: red', 'starts as red');

      this.set('colorValue', 'blue');

      assert.equal(this.element.querySelector('div').getAttribute('style'), 'color: blue', 'updates to blue');
    });
  });
 ```
  
 /cc: @rwjblue @dgeb @buschtoens

**Comment by richgt** [comment]
Looks like we need this solution for our upgrade to Ember 3.24, due to some changes in `linkToExternal` now breaking with the engines not actually set up. Is there something I can do to help push this through and get it committed?

**Comment by villander** [comment]
@richgt we need some effort to land this PR first - https://github.com/emberjs/ember-test-helpers/pull/795

Then, finally, we can release this one to Ember Engines. I'm very excited to see it done and make Engines more stable. I'm waiting for @rwjblue to have some time free to jump on that.

**Comment by rwjblue** [comment]
I've rebased and updated to the new APIs from @ember/test-helpers. I still need to bump to the latest version (separately from this PR).

**Comment by villander** [comment]
@runspired could you please offer some support here?

**Comment by btecu** [comment]
@villander there are some comments that have not been addressed that @bertdeblock left.

**Comment by villander** [comment]
@bertdeblock @btecu I addressed all comments, thank you! 

It still breaking because we need `@ember/test-helpers: 2.5.0` and the current `ember-qunit` [still using 2.4.2](https://github.com/emberjs/ember-qunit/blob/master/package.json#L46)

Also, we need to upgrade all tests here to upgrade the qunit, since we're still using 4.x version

cc: @rwjblue @runspired @SergeAstapov

**Comment by rwjblue** [comment]
I think we are ready to rebase to pull in the changes from #797

**Comment by rwjblue** [comment]
Merged master to get a new set of CI runs, and bring in Ember 4 compat stuff.

File: packages/ember-engines/addon-test-support/index.js [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+  return engineInstance;
+}
+
+async function loadEngine(engineName) {
+  let { owner } = getContext();
+  if (!ownerHasEngine(owner, engineName)) {
+    // ensure that the assets are fully loaded
+    let assetLoader = owner.lookup('service:asset-loader');
+
+    await assetLoader.loadBundle(name);
```
SergeAstapov, line 56: @villander tests fail with `Assertion Failed: No bundle with name "" exists in the asset manifest.` error. I think this line should use `engineName` instead of `name`

villander: Thanks @SergeAstapov it's fixed. Now we need ember-qunit upgrade to get the CI green here.

SergeAstapov: @villander it was done already

villander: It's not true @SergeAstapov the last `ember-qunit` is using the `@ember/test-helpers` 2.4.2, but I and @rwjblue r[eleased a 2.5.0](https://github.com/emberjs/ember-test-helpers/releases/tag/v2.5.0) toat enable ember-engines to run integration tests.

That said we need to merge this PR on `ember-qunit` repo and bum ember engines when it be released - https://github.com/emberjs/ember-qunit/pull/911

SergeAstapov: @villander we have `"@ember/test-helpers": "^2.5.0"` listed in `ember-engines`, see https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/package.json#L40

https://github.com/emberjs/ember-qunit/pull/911 is not needed - it's just a bump of `devDependency` in the `ember-qunit`.

villander: Ok, that's makes sense @SergeAstapov but I'm curious why it's not working 🤔

villander: @SergeAstapov the `ember-qunit` is responsible for exporting `setupRenderingTest`, not the test-helpers package 😄 

https://github.com/emberjs/ember-qunit/blob/472036008c90a8a504907626460a20601a8be82d/addon-test-support/index.js#L56

SergeAstapov:
Closes #525

packages/ember-engines/tests/integration/engine-resolver-test.js
```diff
@@ -1,19 +1,19 @@
-import { module, test } from "qunit";
-import { setupRenderingTest } from "ember-qunit";
-import { render } from "@ember/test-helpers";
-import hbs from "htmlbars-inline-precompile";
-import engineResolverFor from "ember-engines/test-support/engine-resolver-for";
+// import { module, test } from "qunit";
```
SergeAstapov, line 1: @villander should we simply delete this file or uncomment this?

villander: I have to uncomment these lines since we are going to release this API, but we don't want to remove the old ones yet. We going to delete the older API progressively once we migrate our internal tests here. Does it make sense?

packages/ember-engines/addon-test-support/index.js
```diff
@@ -0,0 +1,65 @@
...
+ * @method setupEngine
+ * @param {NestedHooks} hooks
+ * @param {String} engineName
+ * @public
+ */
+
+export function setupEngine(hooks, engineName) {
+  hooks.beforeEach(async function() {
+    if (this.engine !== undefined) {
+      throw new Error('You cannot use `setupEngine` twice for the same test setup. If you need to setup multiple engines, use `loadEngine` directly.');
```
SergeAstapov, line 20: @villander I think this comment misleading as `loadEngine` is not exported?

packages/ember-blog/addon/components/pretty-collor.js
```diff
@@ -0,0 +1,4 @@
+import Component from '@glimmer/component';
+
+export default class PrettyColorComponent extends Component {
```
SergeAstapov, line 3: minor, as this is glimmer component and component class has no behavior - we can drop this file

packages/ember-engines/addon-test-support/index.js
villander: No, it says that you can use `setupEngine` twice in the same test, because we don't have to register `this.engine` twice.

villander: I got your concern, I'll export the `loadEngine`

SergeAstapov (approved):

villander:
@SergeAstapov @rwjblue @runspired can we get it merged and released?

SergeAstapov:
@villander merged!
@rwjblue could you please release to npm?

Gorzas:
The information in this thread wasn't added into the [official documentation](https://ember-engines.com/docs/testing-integration), was it? Let me know in case I'm missing something, open to add the new test API into it 😄.

villander:
@Gorzas that's true buddy. Feel free to open a PR to the documentation repository and I can support you on it 😄

PR #661, opened 2019-06-19
Update deps that were blocked on dropping Node 6.

rwjblue:


rwjblue:
FYI - Since these were all development dependencies, I'm not going to worry about making a release for them...

PR #663, opened 2019-06-20
Migrate Ember Tests to new API

villander:
cc: @rwjblue @dgeb


villander:
@dgeb can we merge this?

PR #669, opened 2019-07-04
[WIP] Adds EngineRouterService to Engines

### LINKED ISSUE #587: Bring the router service to engines
Would be really cool, if we could bring the `RouterService` to engines as well. I propose keeping the exact same API, but adding an `*External` variant for all methods, just like we are doing with the route extension:

https://github.com/ember-engines/ember-engines/blob/ec4d1ae7a413a7e5d9e57a4e3b2e0f0d19a0afcd/addon/-private/route-ext.js#L14-L17

Implementing a proper `RouterService` for engines would also allow us to refactor the `route-ext` and also add the missing `*External` methods to `Route` and `Controller` for symmetry.

I've done a PoC service which I am using for my projects, but it would be great to have an official solution that does not rely on dirty hacks.

#### Links

- [Ember API docs for `RouterService`](https://www.emberjs.com/api/ember/3.4/classes/RouterService)
- [RFC #95: Router Service](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md)
- [PoC implementation using some private API](https://gist.github.com/buschtoens/de2faacfc8977e39843b7e2cc2f3fd1e)

villander:
This fixes #587

### DIFF
--- a/addon/-private/engine-ext.js
+++ b/addon/-private/engine-ext.js
@@ -1,6 +1,7 @@
 import Application from '@ember/application';
 import EngineScopedLinkComponent from '../components/link-to-component';
 import ExternalLinkComponent from '../components/link-to-external-component';
+import EngineRouterService from '../services/engine-router-service';
 import Engine from '@ember/engine';
 
 Engine.reopen({
@@ -10,6 +11,7 @@ Engine.reopen({
     if (!(this instanceof Application)) {
       registry.register('component:link-to', EngineScopedLinkComponent);
       registry.register('component:link-to-external', ExternalLinkComponent);
+      registry.register('service:engineRouterService', EngineRouterService);
     }
 
     return registry;
--- a/addon/services/engine-router-service.js
+++ b/addon/services/engine-router-service.js
@@ -1,0 +1,121 @@
+import Ember from 'ember';
+import Service from '@ember/service';
+import { assert } from '@ember/debug';
+import { computed } from '@ember/object';
+import { reads } from '@ember/object/computed';
+import { getOwner } from '@ember/application';
+
+// @HACK: since `getEngineParent` is not exported
+function getEngineParent(engine) {
+  const symbolPrefix = `__ENGINE_PARENT${Ember.GUID_KEY}`;
+  const symbol = Object.keys(engine).find(k => k.startsWith(symbolPrefix));
+  if (!symbol) {
+    return null;
+  }
+  return engine[symbol];
+}
+
+export default Service.extend({
+  engine: computed(function () {
+    return getOwner(this);
+  }),
+
+  externalRoutes: reads('engine._externalRoutes'),
+
+  mountPoint: reads('engine.mountPoint'),
+
+  rootURL: reads('externalRouter.rootURL'),
+
+  currentURL: reads('externalRouter.currentURL'),
+
+  currentRouteName: computed('externalRouter.currentRouteName', function () {
+    if (this.externalRouter.currentRouteName === this.mountPoint) {
+      return 'application';
+    }
+    return this.externalRouter.currentRouteName.slice(this.mountPoint.length + 1);
+  }),
+
+  externalRouter: computed(function () {
+    return this.rootApplication.lookup('service:router');
+  }),
+
+  rootApplication: computed('engine', function () {
+    let parent = getEngineParent(this.engine);
+    while (getEngineParent(parent)) {
+      parent = getEngineParent(parent);
+    }
+    return parent;
+  }),
+
+  getInternalRouteName(internalRouteName) {
+    // https://github.com/ember-engines/ember-engines/blob/ec4d1ae7a413a7e5d9e57a4e3b2e0f0d19a0afcd/addon/components/link-to-component.js#L52-L57
+    if (internalRouteName === 'application') {
+      return this.mountPoint;
+    }
+    return `${this.mountPoint}.${internalRouteName}`;
+  },
+
+  getExternalRouteName(externalRouteName) {
+    assert(
+      `External route '${externalRouteName}' is unknown.`,
+      externalRouteName in this.externalRoutes
+    );
+    return this.externalRoutes[externalRouteName];
+  },
+
+  transitionTo(routeName, ...args) {
+    return this.externalRouter.transitionTo(
+      this.getInternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  transitionToExternal(routeName, ...args) {
+    return this.externalRouter.transitionTo(
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  replaceWith(routeName, ...args) {
+    return this.externalRouter.replaceWith(
+      this.getInternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  replaceWithExternal(routeName, ...args) {
+    return this.externalRouter.replaceWith(
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  urlFor(routeName, ...args) {
+    return this.externalRouter.urlFor(
+      this.getInternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  urlForExternal(routeName, ...args) {
+    return this.externalRouter.urlFor(
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  isActive(routeName, ...args) {
+    return this.externalRouter.isActive(
+      this.getInternalRouteName(routeName),
+      ...args
+    );
+  },
+
+  isActiveExternal(routeName, ...args) {
+    return this.externalRouter.isActive(
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  }
+});
--- a/tests/helpers/start-app.js
+++ b/tests/helpers/start-app.js
@@ -1,12 +1,12 @@
 import { run } from '@ember/runloop';
-import { merge } from '@ember/polyfills';
+import { assign } from '@ember/polyfills';
 import Application from '../../app';
 import config from '../../config/environment';
 
 export default function startApp(attrs) {
-  let attributes = merge({}, config.APP);
+  let attributes = assign({}, config.APP);
   attributes.autoboot = true;
-  attributes = merge(attributes, attrs); // use defaults, but you can override;
+  attributes = assign(attributes, attrs); // use defaults, but you can override;
 
   return run(() => {
     let application = Application.create(attributes);


addon/-private/engine-ext.js
```diff
@@ -10,6 +11,7 @@ Engine.reopen({
     if (!(this instanceof Application)) {
       registry.register('component:link-to', EngineScopedLinkComponent);
       registry.register('component:link-to-external', ExternalLinkComponent);
+      registry.register('service:engineRouterService', EngineRouterService);
```
rwjblue: We should not need to register under another name here, we should register `service:router` (not a different name)

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+  return engine[symbol];
+}
+
+export default Service.extend({
+  engine: computed(function () {
+    return getOwner(this);
+  }),
+
+  externalRoutes: reads('engine._externalRoutes'),
```
rwjblue: should be underscored, also these are immutable so we should grab them during `init` and store them (not use a CP)

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+
+  mountPoint: reads('engine.mountPoint'),
```
rwjblue: Shouldn't be a CP (grab once and store in `init`) and is private, add `_`

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+  const symbolPrefix = `__ENGINE_PARENT${Ember.GUID_KEY}`;
+  const symbol = Object.keys(engine).find(k => k.startsWith(symbolPrefix));
+  if (!symbol) {
+    return null;
+  }
+  return engine[symbol];
+}
+
+export default Service.extend({
+  engine: computed(function () {
```
rwjblue: Why is this a computed? We should use `getOwner(this)` as needed in each specific location.

Also this name is incorrect for what it does (`getOwner(this)` is going to be an `EngineInstance` instance **not** an `Engine` instance).

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+  currentURL: reads('externalRouter.currentURL'),
+
+  currentRouteName: computed('externalRouter.currentRouteName', function () {
+    if (this.externalRouter.currentRouteName === this.mountPoint) {
+      return 'application';
+    }
+    return this.externalRouter.currentRouteName.slice(this.mountPoint.length + 1);
+  }),
+
+  externalRouter: computed(function () {
```
rwjblue: We should use our existing service aliasing architecture to make the host applications `router` service show up in a different location (e.g. `service:-root-router'`) and then change this to:

```js
_rootRouteService: inject('-root-router'),
```

This will allow us to remove the `getEngineParent` code in this PR AFAICT

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+    return this.rootApplication.lookup('service:router');
+  }),
+
+  rootApplication: computed('engine', function () {
```
rwjblue: Should be removed

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,121 @@
...
+    let parent = getEngineParent(this.engine);
+    while (getEngineParent(parent)) {
+      parent = getEngineParent(parent);
+    }
+    return parent;
+  }),
+
+  getInternalRouteName(internalRouteName) {
+    // https://github.com/ember-engines/ember-engines/blob/ec4d1ae7a413a7e5d9e57a4e3b2e0f0d19a0afcd/addon/components/link-to-component.js#L52-L57
```
rwjblue: The referenced function should be made into a util fucntion like:

```
export function namespaceEngineRouteName(mountPoint, routeName) {
  if (routeName === 'application') {
    return mountPoint;
  } else {
    return `${this.mountPoint}.${routeName}`;
  }
}
```

Then used whereever `getInternalRouteName` is used in this PR (and used in `link-to` at the location this comment points to)

addon/services/engine-router-service.js
villander: @rwjblue I don't get your point sorry, how can I show up this `router` service in a different location?

villander:
@rwjblue can we merge this?

villander:
@buschtoens how to do that? I did your solution, but it doesn't works -https://gist.github.com/buschtoens/de2faacfc8977e39843b7e2cc2f3fd1e#gistcomment-2960117

buschtoens:
I think the missing `routeWillChange` and `routeDidChange` events are not required for the first iteration of this feature. Getting this reduced version the service shipped will already create tremendous value. We may also want to have a broader discussion about the ideal semantics of `routeWillChange` and `routeDidChange` — while I think that my proposal makes sense, some may disagree.

Regarding your comment on my gist: It looks like the event handlers are calling themselves recursively. This can have multiple reasons and I'd need to reproduce it.

villander:
thanks @buschtoens, for while we'll push this without routeWillChange and routeDidChange events. 
let me know when you figure out the recursive problem.

lvegerano:
Will this affect engines that are getting the router service injected into them from the parent app?

cinkonaap:
@lvegerano As the engine router is injected into `service:router`, I'd say yes it will.

However, I feel like this is an anti-pattern, as from how I understand it, Engine routing system shouldn't be concerned with host-app (that's why for example you pass the data via services).

villander:
> Will this affect engines that are getting the router service injected into them from the parent app?

Yes

> However, I feel like this is an anti-pattern, as from how I understand it, Engine routing system shouldn't be concerned with host-app (that's why for example you pass the data via services).

This is not a anti-pattern, we just providing the same API that already exist on router to controllers, components and etc using a service. Nothing changes about isolation/concern. You can check this on my changes on this PR

addon/services/engine-router-service.js
villander: fixed

addon/services/engine-router-service.js
villander: solved

villander:
@rwjblue can you review it again, please?

villander:
For while you guys can use this addon until this PR be approved and released - https://www.npmjs.com/package/ember-engines-router-service

villander:
@rwjblue can you review it again, please?

lvegerano:
@villander 
> Will this affect engines that are getting the router service injected into them from the parent app?
> > Yes

Could you explain (if you can) the type of effect this could have?

villander:
You shouldn't inject the `router` from parent to child engines. It's a bad pattern today, but the only way to inject the `router` service into components/services into engines.

With this solution here it's not needed anymore ; ) 

As mentioned before by my last comment: "**This is not a anti-pattern, we just providing the same API that already exist on router to controllers, components and etc using a service. Nothing changes about isolation/concern. You can check this on my changes on this PR**"

rwjblue:
@villander ya, I'll try to review again this weekend

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+  @reads('externalRouter.rootURL') rootURL;
+
+  @reads('externalRouter.currentURL') currentURL;
+
+  @computed('externalRouter.currentRouteName')
+  get currentRouteName() {
+    if (get(this, 'externalRouter').currentRouteName === this._mountPoint) {
+      return 'application';
+    }
+    return get(this, 'externalRouter').currentRouteName.slice(this._mountPoint.length + 1);
```
buschtoens, line 48: This should probably throw an assertion error, if the current route does not start with the mount point.

addon/-private/engine-ext.js
```diff
@@ -10,6 +11,7 @@ Engine.reopen({
     if (!(this instanceof Application)) {
       registry.register('component:link-to', EngineScopedLinkComponent);
       registry.register('component:link-to-external', ExternalLinkComponent);
+      registry.register('service:router', EngineRouterService);
```
buschtoens, line 14: We should add an assertion / warning for apps that already inject a `router` service.

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+    this.trigger('routeDidChange', ...args);
+  }
+
+  @reads('externalRouter.rootURL') rootURL;
+
+  @reads('externalRouter.currentURL') currentURL;
+
+  @computed('externalRouter.currentRouteName')
+  get currentRouteName() {
+    if (get(this, 'externalRouter').currentRouteName === this._mountPoint) {
```
buschtoens, line 45: You can use a native accessor for `externalRouter`, like you do above already, since it's just a plain getter.

```suggestion
    if (this.externalRouter.currentRouteName === this._mountPoint) {
```

addon/services/engine-router-service.js
buschtoens:
```suggestion
    return this.externalRouter.currentRouteName.slice(this._mountPoint.length + 1);
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+  getExternalRouteName(externalRouteName) {
+    assert(
+      `External route '${externalRouteName}' is unknown.`,
+      externalRouteName in this._externalRoutes
+    );
+    return this._externalRoutes[externalRouteName];
+  }
+
+  transitionTo(routeName, ...args) {
+    return get(this, 'externalRouter').transitionTo(
```
buschtoens, line 64:
```suggestion
    return this.externalRouter.transitionTo(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      namespaceEngineRouteName(this._mountPoint, routeName),
+      ...args
+    );
+  }
+
+  transitionToExternal(routeName, ...args) {
+    return get(this, 'externalRouter').transitionTo(
```
buschtoens, line 71:
```suggestion
    return this.externalRouter.transitionTo(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  }
+
+  replaceWith(routeName, ...args) {
+    return get(this, 'externalRouter').replaceWith(
```
buschtoens, line 78:
```suggestion
    return this.externalRouter.replaceWith(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      namespaceEngineRouteName(this._mountPoint, routeName),
+      ...args
+    );
+  }
+
+  replaceWithExternal(routeName, ...args) {
+    return get(this, 'externalRouter').replaceWith(
```
buschtoens, line 85:
```suggestion
    return this.externalRouter.replaceWith(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  }
+
+  urlFor(routeName, ...args) {
+    return get(this, 'externalRouter').urlFor(
```
buschtoens, line 92:
```suggestion
    return this.externalRouter.urlFor(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      namespaceEngineRouteName(this._mountPoint, routeName),
+      ...args
+    );
+  }
+
+  urlForExternal(routeName, ...args) {
+    return get(this, 'externalRouter').urlFor(
```
buschtoens, line 99:
```suggestion
    return this.externalRouter.urlFor(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      this.getExternalRouteName(routeName),
+      ...args
+    );
+  }
+
+  isActive(routeName, ...args) {
+    return get(this, 'externalRouter').isActive(
```
buschtoens, line 106:
```suggestion
    return this.externalRouter.isActive(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+      namespaceEngineRouteName(this._mountPoint, routeName),
+      ...args
+    );
+  }
+
+  isActiveExternal(routeName, ...args) {
+    return get(this, 'externalRouter').isActive(
```
buschtoens, line 113:
```suggestion
    return this.externalRouter.isActive(
```

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+  get currentRouteName() {
+    if (get(this, 'externalRouter').currentRouteName === this._mountPoint) {
+      return 'application';
+    }
+    return get(this, 'externalRouter').currentRouteName.slice(this._mountPoint.length + 1);
+  }
+
+  get externalRouter() {
+    return this.rootApplication.lookup('service:router');
+  }
```
buschtoens, line 53: I don't think this is solved like suggested in https://github.com/ember-engines/ember-engines/pull/669#discussion_r309728284 yet.

buschtoens: I think @rwjblue was referring to this code specifically:

https://github.com/emberjs/ember.js/blob/08e4da20ce49c5cd58fd9c8113624409fa7bd4ec/packages/%40ember/engine/index.js#L492-L529

buschtoens:
The `routeDidChange` / `routeWillChange` events should probably be marked as experimental / unstable, while a resolution to https://github.com/ember-engines/ember-engines/pull/669#pullrequestreview-294941553 / https://github.com/ember-engines/ember-engines/pull/669#issuecomment-536922027 is pending.

addon/-private/engine-ext.js
villander: I don't think so, with this landing new docs coming up to make explicit that every engine has a properly `router` injected.

addon/services/engine-router-service.js
villander: That's what I did based on `cloneParentDependencies` that has `router:main` from host app. Then I resolved using `router:main` there's only one instance in the ember app and it doesn't exist on engines. I consulted @rwjblue before implement it. 

https://github.com/ember-engines/ember-engines/pull/669/files#diff-ef6a68c610c2dff0c5f5d27e570c8f73

addon/services/engine-router-service.js
villander: when `this._mountPoint` will be `null`? I can't see this scenario since the `mountPoint` is always required to engines - https://github.com/emberjs/ember.js/blob/59f1d40ada46475eff686ca38e93407e533273ee/packages/%40ember/-internals/glimmer/lib/component-managers/outlet.ts#L105

addon/-private/engine-ext.js
buschtoens: Exactly. You're not supposed to override framework services. However prior to this PR you could just inject / register your own `router` service. When these pre-existing apps upgrade, they should get an obvious error / warning, so that they can fix it.

villander: I think it's not our fault, the core team never recommended the users to use `router` service. Then I'd like to skip it for while and release it as soon as possible. Since we have this big lack around the engines ecosystem.

villander:
Hey, @rwjblue I'm looking forward to seeing it merged and released. do you have time to review it until the end of June? The community is expecting a lot for it.

elwayman02:
@rwjblue I believe I need this for the internal projects I've been working on, since we depend heavily on Engines. I'm going to try to unblock myself with the transitionary addon that provides this service in the meantime, though.

addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+
+  @action
+  onRouteWillChange(...args) {
+    this.trigger('routeWillChange', ...args);
+  }
+
+  @action
+  onRouteDidChange(...args) {
+    this.trigger('routeDidChange', ...args);
```
dnachev, line 36: Should these be translated as they will contain the application-wide names instead of the engine-scoped ones (if I'm not mistaken)?

villander:
Rebased & CI green 🚀 

cc: @dgeb @rwjblue

packages/ember-engines/addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+    this.externalRouter.on('routeWillChange', this.onRouteWillChange);
+    this.externalRouter.on('routeDidChange', this.onRouteDidChange);
+  }
+
+  destroy() {
+    this.externalRouter.off('routeWillChange', this.onRouteWillChange);
+    this.externalRouter.off('routeDidChange', this.onRouteDidChange);
+
+    super.destroy();
+  }
```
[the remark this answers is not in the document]
dgeb: >  why do you think it?

@villander typically [`willDestroy`](https://api.emberjs.com/ember/3.20/classes/Service/methods/destroy?anchor=willDestroy) is overridden to implement teardown code. It's registered as a destructor in CoreObject [here](https://github.com/emberjs/ember.js/blob/v3.20.2/packages/%40ember/-internals/runtime/lib/system/core_object.js#L282)

This service lifecycle will be managed by the container, which will call [`destroy`](https://github.com/emberjs/ember.js/blob/v3.20.2/packages/@ember/-internals/container/lib/container.ts#L508) on teardown, so I don't think there's an effective difference in this case. It's just a matter of convention.

> Or even go straight to ember-destroyable-polyfill.

@buschtoens yeah, I would definitely recommend that if we were creating a standard ES class and weren't dealing with an `EmberObject` extension. I'm not sure it's worth the complexity of introducing it here, but that seems fine too.

Ah, so many paths to destruction 🔥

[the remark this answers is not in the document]
villander: Done!

dgeb:
@villander getting close imo! I think we're down to needing a couple tests for the events above, and my main concern which is:

> One of my major concerns here is handling the case when a user explicitly passes the parent router service to the engine. Seems like we should guard against the scenario when a service named router is passed.

We should probably have a hard error when a service named `router` is passed to avoid any potential confusion, and we'll need a version bump. BTW I think it's fine if the host's `router` service is passed but then aliased (say, as `hostRouter`).

packages/ember-engines/addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+
+  @action
+  onRouteWillChange(...args) {
+    this.trigger('routeWillChange', ...args);
```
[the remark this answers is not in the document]
villander: Done!

packages/ember-engines/addon/services/engine-router-service.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @action
+  onRouteDidChange(...args) {
+    this.trigger('routeDidChange', ...args);
```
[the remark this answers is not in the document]
villander: Done!

villander:
@dgeb thanks for bringing all concepts here. 

I added all tests to **EngineRouterService** events, also I added the proper assert to avoid parent pass service named `router` to engines. Then they can pass using other names ; ) 

I think finally we're ready to move it forward and release it as _v0.9.0_

dgeb (approved):
@villander thanks for making those changes!

@rwjblue do you have time to do a final pass reviewing this?

villander:
@rwjblue any change to review it this week?

dgeb:
This PR has dragged on for quite a while and I'd like to move it forward ASAP. That said, I was just doing another walkthrough here and could use some feedback from others in one area ...

I'd like to decide whether we really want to support all the `*External` methods on `EngineRouterService` (e.g. `urlForExternal`, `isActiveExternal`, etc.) We could instead tighten the API surface area by eliminating these methods and instead recommend direct usage of `externalRouter` on `EngineRouterService` (`router.externalRouter.urlFor`, `router.externalRouter.isActive`, etc.). With that said, it might be preferable to shorten `externalRouter` to just `external` to be more concise (`router.external.urlFor`, `router.external.isActive`, etc.)

@villander @rwjblue @buschtoens thoughts?

villander:
> This PR has dragged on for quite a while and I'd like to move it forward ASAP.

Yes, we're fighting with it for one year, and finally, we going to stabilize the Router on Engines. 

> That said, I was just doing another walkthrough here and could use some feedback from others in one area ...

I agree we need more core team involvement here, this has been going on for years.

> I'd like to decide whether we really want to support all the *External methods on EngineRouterService (e.g. urlForExternal, isActiveExternal, etc.) We could instead tighten the API surface area by eliminating these methods and instead recommend direct usage of externalRouter on EngineRouterService (router.externalRouter.urlFor, router.externalRouter.isActive, etc.). With that said, it might be preferable to shorten externalRouter to just external to be more concise (router.external.urlFor, router.external.isActive, etc.)

Seems a great decision to make Ember Engines more ergonomic with default methods from a normal Ember app and easy to update with _RouterService_ coming from Ember.js repository, but we need to treat ember engines seriously and create an RFC for that, can you do that @dgeb?

buschtoens:
@dgeb I'm generally 👍 on your suggestion of removing `*External` methods in favor of an `external` pseudo-router. Having a consistent interface for engine-internal and engine-external routing would make adoption in addons and user-land code slightly less of a pain.

The name `external` / `externalRouter` _could_ be confusing though. One might expect to find the host application `RouterService`.

I still believe that merging the internal and external namespaces as I suggested in #691 is the superior solution in terms of dev ergonomics. Another suggestion was using [opaque route name via a `{{external-route}}` helper](https://github.com/ember-engines/ember-engines/issues/691#issuecomment-612015117).

That said, `router.external` is something I could see myself using much rather than `*External` methods, even if not ideal IMO.

dgeb:
Thanks for your thoughts, @villander and @buschtoens.

> Looking forward to seeing it merged and released.

@villander Same here, but my hesitation is that this PR creates a whole lot of API surface area that is not strictly necessary. I'd like to push forward with the simplest possible version of this PR that provides the minimal API to ship an `EngineRouterService`.

> The name `external` / `externalRouter` could be confusing though. One might expect to find the host application RouterService.

I think you're right here, @buschtoens. Since we're really talking about the `router` service on the _root_ application, it would probably be more accurate to just call this router `root`. So you could access this root router as `router.root.urlFor` for instance.

My proposal to ship this PR as an MVP:

* Rename `EngineRouterService#externalRouter` to `EngineRouterService#root`
* Remove all the `*External` methods from this PR.

Does anyone have an issue with this as a starting point?

This can be followed quickly with a PR that deprecates the no longer needed `*External` methods on controllers and routes. Note that this is inline with the proposal to remove transition methods from controllers and routes in ember core (https://github.com/emberjs/rfcs/pull/674).

We can also consider #691 more carefully after doing all this, since none of this will foreclose on that approach.

buschtoens:
I think we're mixing up two things here. What's currently injected as `externalRouter` is, as you pointed out, the host application `root` router. Calling it `root` makes sense IMO. However, exposing it as public API may not be a good move, at least initially, as this breaks the engine encapsulation principle.

The `*External` methods do not allow to link to arbitrary host app routes. They only allow you to link to host routes that have been explicitly exposed to the engine via the [`dependencies.externalRoutes` config](https://ember-engines.com/docs/links#external-routes).

> My proposal to ship this PR as an MVP:
> 
> * Rename `EngineRouterService#externalRouter` to `EngineRouterService#root`
> * Remove all the `*External` methods from this PR.
>
> This can be followed quickly with a PR that deprecates the no longer needed `*External` methods on controllers and routes.

By doing _just this_ we would break support for `dependencies.externalRoutes`.

So going by your suggestion what we'd need is:

- `root`: the host application `RouterService`, potentially (for now) as a private injection.
- `external` (for lack of a better name): a service / object that implements the `RouterService` interface (or a subset thereof) and allows linking to external routes (host app routes exposed via `dependencies.externalRoutes`).

The advantage of having this `external` `RouterService`-compliant object is that it's _realtively_ pain-free for user-land code and addon code to access `this.router.external` instead of `this.router`, when e.g. the user passes an `@external={{true}}` argument. It would be more complicated to access differently named methods (`*External`) like currently.

I'd like to reiterate this point once more, as it seems that there was some confusion regarding the terminology of host routes (routes accessible via `root`) and external routes (routes exposed via `dependencies.externalRoutes` and accessible via `external` as aliases to host routes):

> I still believe that merging the internal and external namespaces as I suggested in #691 is the superior solution in terms of dev ergonomics. Another suggestion was using [opaque route name via a `{{external-route}}` helper](https://github.com/ember-engines/ember-engines/issues/691#issuecomment-612015117).

villander:
> However, exposing it as public API may not be a good move, at least initially, as this breaks the engine encapsulation principle

I totally agree, it breaks the isolation principle even doing developer ergonomics easier removing surface API. 

Also, I think we should open an RFC to discuss it better there are many pros and cons here to be discussed. We already discussed some of them on the [previous RFC](https://github.com/emberjs/rfcs/pull/122#issuecomment-187245591).

> I still believe that merging the internal and external namespaces as I suggested in #691 is the superior solution in terms of dev ergonomics. Another suggestion was using opaque route name via a {{external-route}} helper.

**I see one big pro:**

1. No new API surface area, other than the external route declaration itself

**I see 4 big trade-offs:**

1. non-obvious what engine and route we're linking
2. non-obvious when reading a route name whether it's internal or external
3. non-obvious to know where the route is defined as well (routes.js or engine.js).
4. Possibility of collisions mentioned by @gabrielcsapo - https://github.com/ember-engines/ember-engines/issues/691#issuecomment-611031219

buschtoens:
> **I see 4 big trade-offs:**
> 
> 1. non-obvious what engine and route we're linking
> 2. non-obvious when reading a route name whether it's internal or external
> 3. non-obvious to know where the route is defined as well (routes.js or engine.js).
> 4. Possibility of collisions mentioned by @gabrielcsapo - [#691 (comment)](https://github.com/ember-engines/ember-engines/issues/691#issuecomment-611031219)

FWIW I addressed these concerns here: https://github.com/ember-engines/ember-engines/issues/691#issuecomment-611411931

villander:
FWIW, I replied to all of them here, addressing all objections  https://github.com/ember-engines/ember-engines/issues/691#issuecomment-611474529 & https://github.com/ember-engines/ember-engines/issues/691#issuecomment-612107078

dgeb:
> By doing just this we would break support for dependencies.externalRoutes.

Apologies. It's been so long since I've used engines personally that I completely overlooked current support for [external routes](https://ember-engines.com/docs/links#external-routes). I will get back up to speed here and work through everyone's comments.

villander:
@dgeb any update or forecast here? Is there something else I can do to move it forward?

btecu:
Bump. This would be really nice to have.

villander:
@btecu you can use this addon that provides this PR to Engines - https://github.com/villander/ember-engines-router-service

PR #688, opened 2019-12-06
shouldCompactReexports returns false for Embroider

tmquinn:


lib/utils/should-compact-reexports.js
```diff
@@ -16,7 +16,8 @@ module.exports = function shouldCompactReexports(addon, options = {}) {
     return (
       checker.for('ember-cli', 'npm').gte('2.13.0') &&
       checker.for('ember-cli-babel', 'npm').gte('6.0.0') &&
-      checker.for('loader.js', 'npm').gte('4.4.0')
+      checker.for('loader.js', 'npm').gte('4.4.0') &&
+      !checker.for('embroider').exists()
```
rwjblue: Hmm, just because embroider exists as a dep doesn’t mean it is being used for the current build. For example, I expect some applications to add embroider as a dependency and have a conditional in their `ember-cli-build.js` while they figure out any issues.

stefanpenner: I actually had this as pending review but apparently forgot to actually submit... anyways I second this concern.

tmquinn: good point

PR #698, opened 2020-03-24
Update project to ember-cli 3.17.0

villander:
Try fix CI related with https://github.com/emberjs/ember.js/pull/18831


dgeb:
@villander thanks for getting this started!

Engines used to be tested back to 2.18. This changes the earliest tested release 3.12. 

@rwjblue do you have an opinion about what you'd like the support matrix to be for ember-engines?

villander:
@dgeb I think we should skip support to 2.18 and just worry about 3.12 and bigger like core team already doing - https://emberjs.com/releases/lts/

dgeb:
@villander this is my preference as well.

villander:
@dgeb can we go ahead and merge it? It's blocking all work around engines

noahgelman:
Is ember engines only supported till 3.12? Is it not supported in octane yet?

villander:
@noahgelman Where did you get that from? Ember Engines has support to ember-cli 3.12 and bigger. And off course to Octane

noahgelman:
@villander None of the documentation has been updated with the newer versions of ember. See my issue here: https://github.com/ember-engines/ember-engines/issues/704

And the examples in the README.md are years old using ember-engines 5.8 so they're not helpful in setting up a modern example using the current version.

PR #707, opened 2020-04-28
Include nested config meta + new default

thoov:
Nested engines config were not being included as meta tags + instead
of throwing an error when the meta tags are not included we default to
an empty pojo.

cc: @rwjblue

### DIFF
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1035,17 +1035,13 @@ module.exports = {
       if (type === 'head') {
         let engineConfig = this.engineConfig(config.environment, {});
         let childrenConfig = this.eachAddonInvoke('contentFor', [type, config]);
+        let content = [];
 
-        let content =
-          '<meta name="' +
-          options.name +
-          '/config/environment" ' +
-          'content="' +
-          escape(JSON.stringify(engineConfig)) +
-          '" />';
+        if (engineConfig === {}) {
+          content.push(`<meta name="${options.name}'/config/environment" content="${escape(JSON.stringify(engineConfig))}" />`);
+        }
 
-        childrenConfig.push(content);
-        return childrenConfig.join('\n');
+        return content.concat(childrenConfig).join('\n');
       }
 
       return '';
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1034,6 +1034,7 @@ module.exports = {
     options.contentFor = function(type, config) {
       if (type === 'head') {
         let engineConfig = this.engineConfig(config.environment, {});
+        let childrenConfig = this.eachAddonInvoke('contentFor', [type, config]);
 
         let content =
           '<meta name="' +
@@ -1043,7 +1044,8 @@ module.exports = {
           escape(JSON.stringify(engineConfig)) +
           '" />';
 
-        return content;
+        childrenConfig.push(content);
+        return childrenConfig.join('\n');
       }
 
       return '';
--- a/lib/engine-addon.js
+++ b/lib/engine-addon.js
@@ -1037,8 +1037,8 @@ module.exports = {
         let childrenConfig = this.eachAddonInvoke('contentFor', [type, config]);
         let content = [];
 
-        if (engineConfig === {}) {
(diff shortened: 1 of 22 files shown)


lib/engine-config-from-meta.js
```diff
@@ -6,7 +6,9 @@ try {
   config = JSON.parse(unescape(rawConfig));
 }
 catch(err) {
-  throw new Error('Could not read config from meta tag with name "' + metaName + '" due to error: ' + err);
+  // in the case where there is no meta tag we can default to an empty pojo
+  // instead of throwing an error
+  config = {};
```
rwjblue, line 11: Lets tweak this to separate the `document.querySelector` from the `JSON.parse(unescape` bit. I'm thinking something like:

```js
var config = {};
var metaElement = document.querySelector('meta[name="{{MODULE_PREFIX}}/config/environment"]');

if (metaElement) {
  try {
    config = JSON.parse(unescape(metaElement.getAttribute('content')));
  } catch (error) {
    throw new Error('Could not read config from meta tag for `{{MODULE_PREFIX}}` due to an error:\n' + error); 
  }
}

export default config;
```

So basically:

* an invalidly composed meta element is still going to be an error (this is important IMHO)
* having no meta element at all is perfectly fine

lib/engine-addon.js
```diff
@@ -1043,7 +1044,8 @@ module.exports = {
           escape(JSON.stringify(engineConfig)) +
           '" />';
 
-        return content;
+        childrenConfig.push(content);
```
rwjblue, line 1047: Can we guard this to avoid pushing the `<meta>` when `engineConfig` is effectively empty (e.g. no reason to push the element)?

lib/engine-config-from-meta.js
thoov: I agree, this is better

lib/engine-addon.js
thoov: Done

lib/engine-addon.js
```diff
@@ -1034,16 +1034,14 @@ module.exports = {
...
-          '/config/environment" ' +
-          'content="' +
-          escape(JSON.stringify(engineConfig)) +
-          '" />';
+        if (Object.keys(engineConfig).length > 0) {
+          content.push(`<meta name="${options.name}/config/environment" content="${escape(JSON.stringify(engineConfig))}" />`);
+        }
 
-        return content;
+        return content.concat(childrenConfig).join('\n');
```
villander, line 1044: missing tests to cover this new scenario

villander:
Thanks @thoov ; )

lib/engine-addon.js
thoov: added!

thoov:
@rwjblue / @villander: tests have been added, this should be good to go.

tests/acceptance/base-nested-engine-test.js
```diff
@@ -0,0 +1,24 @@
+import { module, test } from 'qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+
+module('Acceptance | base + nested engines', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('visiting /base-engine', async function(assert) {
```
villander, line 8: you should write tests to lazyLoading enabled too.

lib/engine-addon.js
villander: @thoov can you also add unit tests in `/node-tests/ultils/engine-addon-test.js` ?

tests/acceptance/base-nested-engine-test.js
villander: @thoov can you also add unit tests in /node-tests/ultils/engine-addon-test.js ? to make sure it will work fine in the Fastboot environment.

villander:
Hey @thoov, any chance to finish this PR to move on?

jonchay:
Hi there! We're seeing this issue as well where we have a nested engine that the top level application doesn't know about. Would this be mergable?

villander:
@thoov could you please rebase this branch with master?

PR #714, opened 2020-06-08
Use default distro (Xenial) instead of Trusty for CI

dgeb:


tests/dummy/lib/common-components/package.json
```diff
@@ -5,7 +5,5 @@
     "ember-addon"
   ],
   "dependencies": {
-    "ember-cli-htmlbars": "*",
```
rwjblue, line 8: Don’t we need these?

dgeb: @rwjblue yeah, I should have marked this PR as draft because I'm in the middle of experimenting.

I removed the deps temporarily to see if they would solve the first issue that CI encounters (see https://travis-ci.org/github/ember-engines/ember-engines/jobs/696185444):

> error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/home/travis/build/ember-engines/ember-engines/tests/dummy/lib/eager-blog/node_modules'". 

Removing the external dummy/lib deps does in fact allow the `yarn install` to complete, but the tests fail as you would expect (see https://travis-ci.org/github/ember-engines/ember-engines/jobs/696234406):

> Addon templates were detected, but there are no template compilers registered for `eager-blog`. Please make sure your template precompiler (commonly `ember-cli-htmlbars`) is listed in `dependencies` (NOT `devDependencies`) in `eager-blog`'s `package.json`.

It seems ideal to use a yarn workspace for this kind of situation, but I don't think that's plausible. Let me know if you have suggestions. This was all working on CI on the last commit to master.

rwjblue: We could use workspaces here if we wanted, mostly to manage these test only deps. For example, ember-auto-import does something like this.

dgeb: @rwjblue do you have an example of a yarn workspace declared in a public project like this?

If I add workspaces at the root I get the error:

> Workspaces can only be enabled in private projects.

I'm assuming that ember-auto-import uses some sort of synthetic private package.json to achieve this effect.

dgeb:
Closing in favor of #719

PR #716, opened 2020-06-22
Add pods structure support for tests

nadavshatz:


addon-test-support/engine-resolver-for.js
```diff
@@ -22,7 +22,8 @@ import EmberResolver from 'ember-resolver';
  */
 export default function engineResolverFor(
   engineName,
-  modulePrefix = engineName
+  modulePrefix = engineName,
+  podModulePrefix = `${engineName}/pods`
```
rwjblue, line 26: I think the change in general is fine, but I think the default should be `undefined`.

nadavshatz: I used the most common value - will it cause issues in case someone uses the classic structure?
I've also considered using the value from `config` but wasn't sure if it's accessible from here, is it?

nadavshatz: @rwjblue ping

if you feel strongly about it - i'll change it and rebase. let me know.

rwjblue: @nadavshatz - I _personally_ never liked using the `/pods` subdir (I always preferred having them directly in the root of `app/`), but my concern is less about what is the most common (or what I like personally) and more about adding a breaking change. Today, calling `engineResolverFor` returns a resolver instance without `podModulePrefix` set; automatically setting that value will change what is resolved by default for all users which therefore changes the thing being tested/exercised by a test; which IMHO is a breaking change.

If you could:

* update to remove the defaulting (or explicitly default it to undefined)
* create a new issue here in this repo to change the default value at the next breaking change release
* leave an inline `// TODO` style comment with a link to that issue

I think we could land this...

nadavshatz: tbh i'm fine with just doing number 1 as well.

rwjblue:
@nadavshatz - The repo structure changed a bit (moved to a monorepo setup) so now there is a bit of a conflict. This file that you are modifying has moved to `packages/ember-engines/addon-test-support/engine-resolver-for.js`, mind fixing up the conflict there?

nadavshatz:
@rwjblue fixed and ready - let me know if there's anything else left to do here.

Cheers 🍻

rwjblue:
Thank you!

PR #730, opened 2020-10-06
Upgrade ember-cli-htmlbars to v5

nlfurniss:


packages/ember-engines/package.json
```diff
@@ -99,7 +99,7 @@
     "calculate-cache-key-for-tree": "^2.0.0",
     "ember-asset-loader": "^0.6.1",
     "ember-cli-babel": "^7.18.0",
-    "ember-cli-htmlbars": "^4.2.3",
+    "ember-cli-htmlbars": "^5.3.1",
```
rwjblue, line 102: For context, 5.x dropped support for Node 8 (which we already dropped)

PR #734, opened 2020-10-21
fix: findHost and findHostHost should allow undefined lazyLoading config

xg-wang:
Normally engine's parent is the project and the findHost loop stops at
the first level dependency engine. If an engine's parent is another
engine it's also fine. However if the parent is an addon that did not
specify lazyLoading the build breaks saying cannot read enabled of
undefined.  This change allows addon uses engines for its dummy app.

### DIFF
--- a/packages/ember-engines/lib/utils/ensure-lazy-loading-hash.js
+++ b/packages/ember-engines/lib/utils/ensure-lazy-loading-hash.js
@@ -10,13 +10,12 @@
 module.exports = function ensureLazyLoadingHash(context) {
   if (
     typeof context.lazyLoading === 'boolean' ||
-    typeof context.lazyLoading === 'undefined'
+    context.lazyLoading === undefined
   ) {
     context.lazyLoading = {
-      enabled: context.lazyLoading,
+      enabled: Boolean(context.lazyLoading),
     };
   }
 
   return context;
 };
-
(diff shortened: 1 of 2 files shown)


packages/ember-engines/lib/utils/ensure-lazy-loading-hash.js
```diff
@@ -8,11 +8,15 @@
  * @return {Object}
  */
 module.exports = function ensureLazyLoadingHash(context) {
-  if (typeof context.lazyLoading === 'boolean') {
+  if (
+    typeof context.lazyLoading === 'boolean' ||
+    typeof context.lazyLoading === 'undefined'
```
rwjblue, line 13: I don't think we need `typeof` here for checking undefined, we can just compare:

```suggestion
    context.lazyLoading === undefined
```

packages/ember-engines/lib/utils/ensure-lazy-loading-hash.js
```diff
@@ -8,11 +8,15 @@
...
+  ) {
     context.lazyLoading = {
-      enabled: context.lazyLoading
+      enabled: context.lazyLoading,
```
rwjblue, line 16: Let's make this a bit more explicit about always being a Boolean:

```suggestion
      enabled: Boolean(context.lazyLoading),
```

rwjblue (changes requested):
Thanks for working on this! Made a couple small suggestions, but otherwise looks good to me!

rwjblue (approved):

xg-wang:
lol forgot to update the test, I blame github apply suggestion too easy to use

rwjblue:
> lol forgot to update the test, I blame github apply suggestion too easy to use

HAHA! yes, I've been trolled by this before as well...

xg-wang:
@rwjblue can we get a release for this change?

Issue #692, opened 2020-02-15
Using <LinkTo> dynamic @route inside engines does not work as expected


abhilashlr:
I've an Ember app (3.16) containing an in-repo addon "ui" and an in-repo engine "admin".
Both the app and the engine share this addon.
One of the components of the addon contains a dynamic `<LinkTo>` usage like below:
```hbs
{{#each @menuItems as |menuItem|}}
  <div class="main-navbar-item">
    {{#if menuItem.model}}
      <LinkTo @route={{menuItem.route}} @model={{menuItem.model}}>
        {{t menuItem.textKey}}
      </LinkTo>
    {{else}}
      {{!-- Tried this approach as well
      {{#link-to menuItem.route}}
        {{t menuItem.textKey}}
      {{/link-to}} --}}
      <LinkTo @route={{menuItem.route}}>
        {{t menuItem.textKey}}
      </LinkTo>
    {{/if}}
  </div>
{{/each}}
```
Using this component inside the host app renders the links well and as expected. While trying to use the same component in the engine and rendering the same, throws the following error:

```
Error: Assertion Failed: You attempted to update `route` on `(unknown object)`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.

`route` was first used:

- While rendering:
  ----------------
    unknownObject.route
...
```

Am I missing something or is this a bug?

roncodes:
Also experiencing the same problem on Octane

roncodes:
I made a pull request, but have a fix in my fork to use in meantime

sbflynn:
Also experiencing this in my engines that use a standalone (ie not in-repo) addon containing a <LinkTo/> element with a dynamic route - blocks our road to Octane.

abhilashlr:
@buschtoens @villander can you please take a look at this issue and the PR made?

chriskrycho:
This dates back to at least 3.12. Here's the outline of the problem as I discovered it:

1. Engines prefix route names by their mount point: [here](https://github.com/ember-engines/ember-engines/blob/d808e6b505f215c32028f210393144468bf8d2d3/addon/components/link-to-component.js#L23).
2. This is applied by calling `set` on the `LinkTo` component: [here](https://github.com/ember-engines/ember-engines/blob/d808e6b505f215c32028f210393144468bf8d2d3/addon/components/link-to-component.js#L52).
[...]

PR #735, opened 2020-10-22
Add mount prefix to route via computed instead of set

elwayman02:
Fixes https://github.com/ember-engines/ember-engines/pull/668
Closes https://github.com/ember-engines/ember-engines/pull/693

https://github.com/ember-engines/ember-engines/pull/693 attempts to fix this same problem by avoiding a call to `set`, but I think that solution is not ideal as it is still adding side effects to `didReceiveAttrs`.

There is further potential for improvement here, but I only fixed the immediate issue I had so as to touch as small a surface as possible. Other possible areas where this fix may need to be replicated:

- `targetRouteName`
- `current-when`

Both of the above properties are ALSO `set` the same way `route` was, so it stands to reason that they may also have this same bug. We could try to fix them all here if you think it wise.

Similarly, `link-to-external` seems to have exactly the same problem and likely *also* needs this fix.

However, I was unable to figure out how to write a test that replicates this error - I took the reproduction from https://github.com/ember-engines/ember-engines/pull/668
I think the only way to make this work in a test is to write a wrapping component that holds the property and yields it to the `link-to` - apparently setting a property in the context of a test does not surface the same behavior protecting against modifications.

I can take a stab at the more complicated test setup tomorrow, but I wanted get this up for review to see if this even looked like the right approach. I have verified it in the demo app and ensured all broken tests were fixed as well.

cc: @rwjblue

### DIFF
--- a/packages/ember-blog/addon/controllers/post.js
+++ b/packages/ember-blog/addon/controllers/post.js
@@ -3,6 +3,7 @@ import { set } from '@ember/object';
 
 export default Controller.extend({
   queryParams: ['lang'],
+  commentsRoute: 'post.comments', // Added to demonstrate that dynamic route names work
 
   actions: {
     transitionToHomeFromController() {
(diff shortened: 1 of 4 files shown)


rwjblue (approved):

villander:
@rwjblue I think this PR also fix the #692

PR #750, opened 2021-01-28
Compute LCA host for engines

brendenpalmer:
This updates the host addons logic for lazy engines to compute the lowest comment ancestor (that's considered a host) for all engines by the same name in the project. This is important because lazy engines only get output on disk to a single location. It's entirely possible for host addons in different contexts to be different, which could potentially cause runtime bugs due to missing dependencies.

For example, given the following project addon structure:

```
        --Project--
         /      \
        /        \
   Lazy Engine A  \
                Addon A
                  |
                  |
             Lazy Engine B
              /          \
             /            \
        Lazy Engine A   Lazy Engine C
```

It wouldn't be expected for Lazy Engine A to use Lazy Engine B (and above) to compute its host addons; when lazy engine A is later loaded by the project at the top-level some addons could be missing at runtime; e.g., those that would have been provided by Lazy Engine B and therefore be deduped in that context.


packages/ember-engines/lib/utils/find-lca-host.js
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+
+    // traverse upwards for each engine/addon
+    equivalentLevels = equivalentLevels.map(({ parent }) => parent);
+  }
+
+  // this should never be triggered
+  throw new Error(
+    `[ember-engines] Could not find a common host for: \`${engineName}\``
```
rwjblue, line 159: Can you update this error message to also include the root paths of all of the engines that we found for the current name?

rwjblue (approved):

Issue #749, opened 2021-01-28
0.8.10 fails because an Embroider macro doesn't support parallel transpilation


chriskrycho:
Error output:

```
ERROR Summary:

  - broccoliBuilderErrorStack: [undefined]
  - code: [undefined]
  - codeFrame: [undefined]
  - errorMessage: [broccoli-persistent-filter:Babel > [Babel: ember-engines]: Babel: ember-engines] was configured to `throwUnlessParallelizable` and was unable to parallelize a plugin. 
plugins:
1: name: unknown, location: unknown

Please see: https://github.com/babel/broccoli-babel-transpiler#parallel-transpilation for more details
  - errorType: [undefined]
  - location:
    - column: [undefined]
    - file: [undefined]
    - line: [undefined]
  - message: [broccoli-persistent-filter:Babel > [Babel: ember-engines]: Babel: ember-engines] was configured to `throwUnlessParallelizable` and was unable to parallelize a plugin. 
plugins:
1: name: unknown, location: unknown
```

rwjblue:
This was introduced by  https://github.com/ember-engines/ember-engines/commit/bdc186ed8f98aad85eb734c26ba5b3ec5dab246b, apparently `@embroider/macros` is not parallelizable.

PR #751, opened 2021-01-28
Ensure ember-engines are parallelizable.

rwjblue:
Replicate https://github.com/ember-engines/ember-engines/issues/749


rwjblue:
Now that #752 has landed, I think this should pass.

PR #758, opened 2021-03-18
Expose a `buildEngine` function (avoid the appearance of being an `Ember.Object`)

bertdeblock:
[More info explaining the why.](https://github.com/ember-engines/engine-blueprint/pull/27#discussion_r593773644)

Best to view the diff without whitespace changes, but basically what I've done is this:

```js
function buildEngine() {}

module.exports = {
  buildEngine,
  extend: buildEngine,
};
```

### DIFF
--- a/packages/ember-engines/node-tests/unit/engine-addon-test.js
+++ b/packages/ember-engines/node-tests/unit/engine-addon-test.js
@@ -10,6 +10,26 @@ const path = require('path');
 const memoize = require('../../lib/utils/memoize');
 
 describe('engine-addon', function() {
+  describe('building an engine', function() {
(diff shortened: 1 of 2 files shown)


packages/ember-engines/node-tests/unit/engine-addon-test.js
```diff
@@ -10,6 +10,26 @@ const path = require('path');
 const memoize = require('../../lib/utils/memoize');
 
 describe('engine-addon', function() {
+  describe('building an engine', function() {
```
bertdeblock, line 13: No idea if this is enough of a test, feedback appreciated!

PR #759, opened 2021-03-18
Support non-camelized engine names (deprecate camelization)

### LINKED ISSUE #625: [RFC (Sorta)] Camelizing Engine Config is Annoying, Let's Not
In parent application's we're asked to build out engine configurations using camelized package names. This is both surprising, and awkward — especially for scoped packages. Surprising because our packages already have names used by NPM/Yarn, e.g. "my-blog", or "my-super-awesome-engine". Awkward for scoped packages, e.g. given "@precision-nutrition/my-engine" becomes "@precisionNutrition/myEngine". Both require me to internalize camelization rules.

I hereby propose that we deprecate camelized engine names in the parent configs in favour of their natural package names.

Before
------

```javascript
const engineConfig = {
  '@precisionNutrition/myEngine': {
    // snip snip snip
  },
};
```

After
-----

```javascript
const engineConfig = {
  '@precision-nutrition/my-engine': {
    // snip snip snip
  },
};
```

Alternatives
=========

Beef up the documentation so that scoped package users aren't spelunking the code to figure out why things aren't working.

bertdeblock:
Closes #625.

I'm guessing some of the deprecation details can be improved (id, message, ...).

### DIFF
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -1,5 +1,5 @@
 import { camelize } from '@ember/string';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import EngineInstance from '@ember/engine/instance';
 
 EngineInstance.reopen({
@@ -58,10 +58,26 @@ EngineInstance.reopen({
     if (!dependencies) {
       dependencies = {};
 
-      let camelizedName = camelize(name);
+      let engineConfigurations = this.base.engines || {};
+      let engineConfigurationKey = name;
+      let engineConfiguration = engineConfigurations[engineConfigurationKey];
 
-      let engineConfiguration =
-        this.base.engines && this.base.engines[camelizedName];
+      if (!engineConfiguration) {
+        engineConfigurationKey = camelize(name);
+        engineConfiguration = engineConfigurations[engineConfigurationKey];
+
+        deprecate(
+          `Support for camelized engine names has been deprecated. Please use '${name}' instead of '${engineConfigurationKey}'.`,
+          typeof engineConfiguration !== 'object',
+          {
+            id: 'ember-engines.deprecation-camelized-engine-names',
+            for: 'ember-engines',
+            until: '0.9.0',
+            since: {
+              enabled: '0.8.13',
+            },
+          });
+      }
 
       if (engineConfiguration) {
         let engineDependencies = engineConfiguration.dependencies;
@@ -88,7 +104,7 @@ EngineInstance.reopen({
                 let dependency = this.lookup(dependencyKey);
 
                 assert(
-                  `Engine parent failed to lookup '${dependencyKey}' dependency, as declared in 'engines.${camelizedName}.dependencies.${category}'.`,
+                  `Engine parent failed to lookup '${dependencyKey}' dependency, as declared in 'engines.${engineConfigurationKey}.dependencies.${category}'.`,
                   dependency
                 );
 
--- a/packages/ember-engines/tests/unit/engine-instance-test.js
+++ b/packages/ember-engines/tests/unit/engine-instance-test.js
@@ -157,3 +157,69 @@ test('it can build a child engine instance with dependencies that are aliased', 
     );
   });
 });
+
+test('it deprecates support for camelized engine names', async function(assert) {
+  assert.expect(6);
+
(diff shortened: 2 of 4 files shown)


packages/ember-engines/tests/unit/engine-instance-test.js
```diff
@@ -157,3 +158,77 @@ test('it can build a child engine instance with dependencies that are aliased',
     );
   });
 });
+
+test('it deprecates camelized engine names', async function(assert) {
+  assert.expect(5);
+
+  let deprecationMessages = [];
+
+  registerDeprecationHandler((message, options, next) => {
```
bertdeblock, line 167: Not sure if this is the correct way. I also tried using `ember-qunit-assert-helpers` but asserting that no deprecation was thrown didn't work for me. The `globals-resolver` deprecation was also thrown during this test and I couldn't manage to filter it out.

packages/ember-engines/addon/-private/engine-instance-ext.js
```diff
@@ -58,10 +58,26 @@ EngineInstance.reopen({
...
+
+        deprecate(
+          `Support for camelized engine names has been deprecated. Please use '${name}' instead of '${engineConfigurationKey}'.`,
+          typeof engineConfiguration !== 'object',
+          {
+            id: 'ember-engines.deprecation-camelized-engine-names',
+            for: 'ember-engines',
+            until: '1.0.0',
+            since: {
+              enabled: '0.8.13',
```
bertdeblock, line 77: Or would this be `0.9.0`?

packages/ember-engines/addon/-private/engine-instance-ext.js
```diff
@@ -58,10 +58,26 @@ EngineInstance.reopen({
...
+        engineConfigurationKey = camelize(name);
+        engineConfiguration = engineConfigurations[engineConfigurationKey];
+
+        deprecate(
+          `Support for camelized engine names has been deprecated. Please use '${name}' instead of '${engineConfigurationKey}'.`,
+          typeof engineConfiguration !== 'object',
+          {
+            id: 'ember-engines.deprecation-camelized-engine-names',
+            for: 'ember-engines',
+            until: '1.0.0',
```
rwjblue, line 75: should be 
```suggestion
            until: '0.9.0',
```

packages/ember-engines/addon/-private/engine-instance-ext.js
rwjblue: 0.8.13 is correct (at the moment)

packages/ember-engines/tests/unit/engine-instance-test.js
rwjblue: I'd do this setup instead:


```js
// packages/ember-engines/tests/test-helper.js

let deprecations;
registerDeprecationHandler((message, options, next) => {
  // in case a deprecation is issued before a test is started
  if (!deprecations) {
    deprecations = [];
  }

  deprecations.push(message);
  next(message, options);
});

QUnit.testStart(function () {
  deprecations = [];
});

QUnit.assert.noDeprecations = function (callback) {
  let originalDeprecations = deprecations;
  deprecations = [];

  callback();
  this.deepEqual(deprecations, [], 'Expected no deprecations during test.');

  deprecations = originalDeprecations;
};

QUnit.assert.deprecations = function (callback, expectedDeprecations) {
  let originalDeprecations = deprecations;
  deprecations = [];

  callback();
  this.deepEqual(deprecations, expectedDeprecations, 'Expected deprecations during test.');

  deprecations = originalDeprecations;
};

QUnit.assert.deprecationsInclude = function (expected) {
  this.pushResult({
    result: deprecations.indexOf(expected) > -1,
    actual: deprecations,
    message: `expected to find \`${expected}\` deprecation`,
  });
};
```

Which is basically what I do over in https://github.com/emberjs/ember-test-helpers/blob/d1a05f4409b58c1fd1689a3afbabcf8f8e4047ab/tests/test-helper.js

rwjblue: The issue with the current approach is that the test essentially mutates global state (there is no way to _undo_ the registerDeprecationHandler).

rwjblue (changes requested):

rwjblue:
Just a few minor issues, but overall I'm very much in favor.

rwjblue (approved):

PR #761, opened 2021-03-25
[BUGFIX LTS] override assertLinkToOrigin method to noop in ember-engines

hmajoros:
This PR extends off the work done in ember.js repo here: https://github.com/emberjs/ember.js/pull/19477/files

since LinkToExternal extends LinkTo, we only want to run the assertion around using LinkTo inside of a routeless engine in the LinkTo class, and not the extension of LinkToExternal.

### DIFF
--- a/packages/ember-engines/addon/components/link-to-external.js
+++ b/packages/ember-engines/addon/components/link-to-external.js
@@ -13,6 +13,10 @@ if (gte('ember-source', '3.24.0-alpha.1')) {
 
       return externalRoute;
     }
+
+    // override LinkTo's assertLinkToOrigin method to noop. In LinkTo, this assertion
+    // checks to make sure LinkTo is not being used inside a routeless engine
+    assertLinkToOrigin() {}
   };
 } else {
   LinkToExternal = LinkComponent.extend({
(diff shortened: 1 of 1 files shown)


packages/ember-engines/addon/components/link-to-external.js
```diff
@@ -13,6 +13,10 @@ if (gte('ember-source', '3.24.0-alpha.1')) {
 
       return externalRoute;
     }
+
+    // override LinkTo's assertLinkToOrigin method to noop. In LinkTo, this assertion
+    // checks to make sure LinkTo is not being used inside a routeless engine
+    assertLinkToOrigin() {}
```
rwjblue, line 19: Can you update this comment to link to the PR over in Ember?

hmajoros: done

chancancode:
This seems fine, and the patch was released in 3.27.0-beta.3 for now, and will make its way to release/LTS eventually. *However*, 3.27.0 and above also deprecated reopening/extending the built-in component, so we should probably find a better way to handle this (we can add another "secret handshake" or just support the feature in Ember directly).

Issue #756, opened 2021-03-09
Prepare for addition of engine-specific router service


dgeb:
Engines pre-date the introduction of the `router` service. Ideally, when that service was introduced, engines should have simultaneously introduced their own engine-specific `router` service to provide restricted access to an engine's own routes. Unfortunately, this was not done, and engines instead unintentionally inherited their host's `router` service, breaking the encapsulation that is a goal of this project.  

@villander has made a lot of progress working on an engine-specific router service in #669. However, because many apps currently make use of the host's `router` service, we can't simply swap in a new `router` service with an entirely different scope without breaking those apps.

Before introducing the new `router` service, we need to deprecate the current `router` service in a semver-compliant way, probably as an 0.8.x release. This will clear the way for complete removal of the `router` service in 0.9.x, followed by a re-introduction of an engine-specific `router` service in a release >= 0.10. 

The initial deprecation work involves the following tasks:

* Register a new `router` service that proxies all its methods to the host's `router` service. Every method invocation that's passed should include a deprecation warning.
* Create a deprecation documentation page about the router service on ember-engines.com. Provide suggestions for how to register the host's router as `hostRouter` and/or the root router as `appRouter`. However, we don't want to automatically create either of these services in an engine - the decision should be the developer's.

We should also consider allowing an experimental build flag to allow usage of an alpha engine-specific router which we can iterate upon starting in 0.9. This approach should allow opt-in usage of a router service similar to that in #669 for developers who are willing to experiment and endure some breaking changes.

This approach was discussed in February's engines team meeting with @rwjblue and @buschtoens. Please let me know if this ticket needs any adjustments / further discussion before we begin the work. @villander has already discussed reusing much of the work in #669 to deprecate the current `router` service.

buschtoens:
Sounds exactly like what I recall from the meeting. 👍🏼 

I have a few minor notes though:

> - Register a new `router` service that proxies all its methods to the host's `router` service. Every method invocation that's passed should include a deprecation warning.

Unconditionally creating a proxy `router` service in every engine may cause unintended side-effects, e.g. some addon or user code may check for the existence of a `router` service and use different adapters then. Considering that addons that deal with routing need to special-handle engines in one way or another, if they want to be compatible, the likelihood is not zero, that the mere presence of a `router` service could change the behavior.

On the other hand, this may be super-edge-case-y, SemVer-lawyering-wise still okay and not even a real issue.

[...]

PR #764, opened 2021-04-08
Add deprecation warning to router service from host

villander:
It creates a deprecation warning to #756 starting the addition of _Engine Router Service._ 

Also, it provides suggestions for how to register the host's router as **hostRouter** and/or the root router as **appRouter**. 

Deprecations page has been linked - https://github.com/ember-engines/ember-engines.com/pull/111

### DIFF
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -1,6 +1,7 @@
 import { camelize } from '@ember/string';
 import { assert, deprecate } from '@ember/debug';
 import EngineInstance from '@ember/engine/instance';
+import { deprecate } from '@ember/debug';
 
 EngineInstance.reopen({
   /**
@@ -102,6 +103,18 @@ EngineInstance.reopen({
 
                 let dependencyKey = `${dependencyType}:${dependencyNameInParent}`;
                 let dependency = this.lookup(dependencyKey);
+
+                deprecate(
+                  `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of '${dependencyName}'.`,
+                  category === 'services' && dependencyName === 'router',
+                  {
+                    id: 'ember-engines.deprecation-router-service-from-host',
+                    for: 'ember-engines',
+                    until: '0.9.0',
+                    since: {
+                      enabled: '0.8.13',
+                    },
+                  });
 
                 assert(
                   `Engine parent failed to lookup '${dependencyKey}' dependency, as declared in 'engines.${engineConfigurationKey}.dependencies.${category}'.`,
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -64,7 +64,9 @@ EngineInstance.reopen({
   },
 
   lookup(fullName, options) {
-    deprecateHostRouterService(fullName);
+    if (fullName === 'router') {
+      deprecateHostRouterService(fullName);
+    }
     return this._super(fullName, options);
   },
 
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -2,9 +2,9 @@ import { camelize } from '@ember/string';
 import { assert, deprecate } from '@ember/debug';
 import EngineInstance from '@ember/engine/instance';
 
-function deprecateHostRouterService(routerName) {
+function deprecateHostRouterService() {
   deprecate(
-    `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of '${routerName}'.`,
+    `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of 'router'.`,
     false,
     {
       id: 'ember-engines.deprecation-router-service-from-host',
@@ -64,8 +64,8 @@ EngineInstance.reopen({
   },
 
   lookup(fullName, options) {
-    if (fullName === 'router') {
-      deprecateHostRouterService(fullName);
+    if (['service:router', 'router:main'].includes(fullName)) {
+      deprecateHostRouterService();
     }
     return this._super(fullName, options);
   },
@@ -126,7 +126,7 @@ EngineInstance.reopen({
                 let dependency = this.lookup(dependencyKey);
 
                 if (category === 'services' && dependencyName === 'router') {
-                  deprecateHostRouterService(dependencyName);
+                  deprecateHostRouterService();
                 }
 
                 assert(
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -13,7 +13,7 @@ function deprecateHostRouterService(routerName) {
       url: 'https://ember-engines.com/docs/deprecations#host-router-service',
       since: {
         enabled: '0.8.13',
-      },
+      }
     });
 }
 
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -63,13 +63,6 @@ EngineInstance.reopen({
     this._externalRoutes = {};
   },
 
-  lookup(fullName, options) {
-    if (['service:router', 'router:main'].includes(fullName)) {
-      deprecateHostRouterService();
-    }
-    return this._super(fullName, options);
-  },
-
   buildChildEngineInstance(name, options = {}) {
     // Check dependencies cached by engine name
     let dependencies =
--- a/packages/ember-engines/addon/-private/engine-instance-ext.js
+++ b/packages/ember-engines/addon/-private/engine-instance-ext.js
@@ -2,6 +2,21 @@ import { camelize } from '@ember/string';
 import { assert, deprecate } from '@ember/debug';
 import EngineInstance from '@ember/engine/instance';
 
+function deprecateHostRouterService(routerName) {
+  deprecate(
+    `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of '${routerName}'.`,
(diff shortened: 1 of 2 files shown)


dgeb (approved):
@villander This is a good start - thank you!

dgeb:
While this PR covers explicit importing of the host's `router` with the name `router`, we also need to deprecate accessing the host's `router` service via `lookup`, since it is unfortunately automatically injected right now. But let's tackle that in a separate PR.

@rwjblue please review when you have a chance.

villander:
@dgeb @rwjblue do you have any idea of how to overwrite this method to deprecate the access via `lookup`? 

https://github.com/emberjs/ember.js/blob/21bd70c773dcc4bfe4883d7943e8a68d203b5bad/packages/%40ember/-internals/container/lib/container.ts#L150-L156

villander:
> we also need to deprecate accessing the host's router service via lookup

@dgeb I added it as well, feel free to review it again ; )

villander:
I ran ember try locally and everything is ok. It seems a CI issue 

**result:**

All 10 scenarios succeeded
✨  Done in 404.39s.

packages/ember-engines/addon/-private/engine-instance-ext.js
```diff
@@ -48,6 +63,11 @@ EngineInstance.reopen({
     this._externalRoutes = {};
   },
 
+  lookup(fullName, options) {
+    deprecateHostRouterService(fullName);
```
bertdeblock, line 67: This now triggers the deprecation warning for _every_ lookup.

Also, in case of doing `lookup('service:router')`, the deprecation message makes it seem like you have to do `lookup('hostRouter')` or `lookup('appRouter')` instead.

Maybe it's better to have separate deprecations/messages for these?

Additionally, in https://github.com/ember-engines/ember-engines/pull/759, a setup was added for testing deprecations.

villander: Thanks a lot! I'll a condition for that, I lost it on refactoring

> Maybe it's better to have separate deprecations/messages for these?

Why we need separate messages? The users are trying to get the app router from the root application directly. The message should be the same, warn them to avoid inject `router` directly without any alias.

bertdeblock: This:
```js
fullName === 'router'
```
Should be:
```js
fullName === 'service:router'
```
`router` is not a full name.

I think the current tests are passing by accident actually. The last two assertions pass because the deprecation was already captured before. If you change `blogEngineInstance.lookup('service:router');` to `blogEngineInstance.lookup('service:blablabla');` the tests still pass. So I think the current test is not reliable.

> Why we need separate messages? The users are trying to get the app router from the root application directly.

Yes, the user requests access to the host's router in both cases, but _the way_ the user does it in both scenario's (defining a dependency vs doing a manual lookup) is different and I feel that using the same message to cover both scenario's isn't clear.

That could just be me of course. I still feel like the current deprecation message doesn't really fit the lookup scenario: `Please use a different name as 'hostRouter' or 'appRouter' instead of 'service:router'.` makes it seem that the user needs to do `lookup('hostRouter')` instead which isn't the case.

villander: > he tests still pass. So I think the current test is not reliable.

I agree, what's happening here is a cache in `assert.deprecationsInclude` the extension on Qunit that you added on #759 has a bug. 

> Yes, the user requests access to the host's router in both cases, but the way the user does it in both scenario's (defining a dependency vs doing a manual lookup) is different and I feel that using the same message to cover both scenario's isn't clear.

I don't agree here! We shouldn't say anything about `lookup` because the first thing to do is declare `router` as an alias like `{ 'app-router': 'router' }`, then we should not suggest to the users use `lookup('hostRouter')` directly jumping the first step (`{ 'app-router': 'router' }`). The Ember.js users already know everything about lookup and how to use that with different names. Then, IMO the deprecate message should be the same, given we're deprecating the same "feature" in different levels.

bertdeblock: It's not a bug 😀. `assert.deprecationsInclude` should **not** change the state during a single test.
It should only check if a deprecation is included or not. Instead, you could extend the method
to check if a deprecation is included X times.

For example (just a suggestion):
```js
assert.deprecationsInclude(deprecationMessage, { count: 2 });
```
Simply removing the other assertions does not make the test reliable again.

**Regarding the lookup:**

Sorry, I have a hard time understanding what you are trying to say. I'll try to explain myself better.

It _is_ possible that the user does a lookup of `service:router` **without** defining the dependency.
As mentioned in [this comment](https://github.com/ember-engines/ember-engines/pull/764#issuecomment-816721788) the router service is injected automatically.
In this scenario I still feel that the current deprecation message is not clear enough.
The guide only covers the case where the user has already defined the dependency.

I am **not** saying we should suggest using `lookup('hostRouter')`.
I am only saying that the current deprecation **makes it seem** like the user should.
Which I feel can be confusing.

Additionally, the deprecation is now also triggered for `router:main`.
**even though the user isn't doing this lookup themselves**.
So there's no way for the user to get rid of the deprecation in this case.
Again, this is because the router service is already injected automatically (I think).
Manually defining the dependency **with or without** an alias won't resolve the deprecation here.
You can test this in the dummy app. The `ember-blog` does not have a dependency on the router service and it also does not do a lookup of the router service, yet the deprecation message is displayed.

villander: > Simply removing the other assertions does not make the test reliable again.

I just removed them because they're not possible to test. And if you put `{ count: 25 }` or any other number it will continue passing. The problem here is because this assertion is not reliable. 


```js
return blogEngineInstance.boot().then(() => {
    blogEngineInstance.lookup('service:router');

    assert.deprecationsInclude(
      `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of 'router'.`,
    { count: 2 });

    // app router from root application
    blogEngineInstance.lookup('router:main');

    assert.deprecationsInclude(
      `Support for the host's router service has been deprecated. Please use a different name as 'hostRouter' or 'appRouter' instead of 'router'.`,
    { count: 7 });
  });
 ```
  
  > Additionally, the deprecation is now also triggered for router:main.
even though the user isn't doing this lookup themselves.

Many users use `router:main` to get the router from host application. Then it's not true!

bertdeblock: I have tested this myself, otherwise I would not suggest it as a possible solution. I can understand that it's a bit harder to test, but it's not impossible. You would have to update the test setup yourself for it to work though.

> Many users use `router:main` to get the router from the host application.

I did not say that there aren't any users that do this.
I said that, at the moment the deprecation is also triggered without looking up `router:main` manually, you can see this when running the dummy app.
Triggering the deprecation for logic over which the user has no control seems strange.
There's also already a [lint rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-private-routing-service.md) for this case, so maybe triggering the deprecation for `router:main` is not needed?

Regarding the lookup, I tested this again and indeed, it seems the lookup returns `undefined` for the router service.
I must have been mistaken the first time I tested it, so you are right.

I'm feeling some frustration on your side regarding my feedback, so I'm backing off as I feel this is adding too much noise to the PR. Thanks for working on this though 👍.

villander: > I'm feeling some frustration on your side regarding my feedback

I'm not frustrated. It's a pleasure for me to teach about Ember Engines. Thank you so much for challenging me to write the tests here and improve the code 😄

villander:
@rwjblue could you please review this work?

dgeb:
If as [this discussion](https://github.com/ember-engines/ember-engines/pull/764#discussion_r611214742) indicates, the host's `router` is NOT being automatically injected, then there is indeed no need to add the deprecation to `lookup` (which just returns `undefined`).

I'm confused because this explicitly came up during our last team meeting in which we discussed adding the deprecation to `lookup` because of this injection. However, if it's not happening, and the only way for the host's `router` service to be accessed is by explicitly injecting it, then it would seem to me that the deprecation in `buildChildEngineInstance` is the only one that is needed.

@rwjblue I _think_ it was you who raised the point about automatic injection, so please let me know if I'm  misunderstanding something.

rwjblue:
@dgeb - The issue isn't that we auto-inject `service:router` from the top level parent into Engines specifically (because as you mentioned, we don't do that). The issue is that the _meaning_ of the APIs on `service:router` from _within_ an Engine today is **the same** as it means from the top level parent (e.g. methods are not properly scoped to the engines mount point). We have to deprecate using `service:router` (even though it definitely gets its own engine local instance today) because if you call `routerService.urlFor(...)` you have to invoke it _including_ the mount point, **but** we _want_ to make it automatically scoped to the mount point. This will be a breaking change, which is why we need this deprecation (to guide folks away from the breakage).

rwjblue:
Hmm, I tried checking where Ember registers `service:router` for lookup and it _seems_ that it is only done for `@ember/application`'s registry, which would infer that within an Engine `@service router` is **not** available unless you manually wire it through (in other words, my comment might be wrong). 

Can someone confirm?

bertdeblock:
@rwjblue Yup, you are right. The `router` service is not available automatically within an engine.

dgeb:
@rwjblue @bertdeblock thanks for confirming.

@villander can you please update this PR to remove the `lookup` deprecation completely? Sorry for all the confusion.

villander:
Yup @rwjblue @dgeb  the `service:router` is not injected automatically, but many users use `router:main` to get the router from the host application. It's automatically injected into every engine. Also, we've been using it on #669

Do we want to deprecate the access to `router:main` into engines?

dgeb (approved):
Thanks @villander - I think this looks good now.

villander:
Could you please merge it @dgeb? I have no privilege to do it

rwjblue (approved):

bertdeblock:
I think we forgot to update `enabled: '0.8.13',`.

dgeb:
@bertdeblock good call - this is corrected in https://github.com/ember-engines/ember-engines/pull/774

PR #769, opened 2021-05-05
Support addon bundle caching as part of `ancestorHostAddons`

brendenpalmer:
This adds support for bundle addon caching as part of `ember-cli`; specifically, we need to access the real/target instance in the event we're seeing a proxy as part of `ancestorHostAddons`. If we don't do this, the host addons won't be correct, since as part of this optimization `ember-cli` returns an empty array for `addons`. The main reason for this is that `ember-engines` dedupe is _stateful_: it's possible for `_orginalAddons` to be clobbered by `addons` again so the real original addons will never be restored as part of `restoreOriginalAddons`, producing a build that's broken/incorrect.

For more information, see the PR adding this to `ember-cli`: https://github.com/ember-cli/ember-cli/pull/9487.

### DIFF
--- a/packages/ember-engines/lib/engine-addon.js
+++ b/packages/ember-engines/lib/engine-addon.js
@@ -44,6 +44,27 @@ const findRoot = require('./utils/find-root');
 const findHost = require('./utils/find-host');
 const processBabel = require('./utils/process-babel');
 
+// support addon bundle caching; we need to use the `TARGET_INSTANCE`
+// symbol to access the real addon when computing `ancestorHostAddons`
+// for more information, see the original PR in `ember-cli` adding this
+// optimization: https://github.com/ember-cli/ember-cli/pull/9487
+let TARGET_INSTANCE_SYMBOL = null;
+
+try {
+  const targetInstanceModule = require('ember-cli/lib/models/per-bundle-addon-cache/target-instance');
+
+  if (targetInstanceModule) {
+    TARGET_INSTANCE_SYMBOL = targetInstanceModule.TARGET_INSTANCE;
+  }
+} catch (e) {
+  // we only want to handle the error when this module isn't found; i.e.,
+  // when a consumer of `ember-engines` is using an old version of `ember-cli`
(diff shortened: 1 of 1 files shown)


packages/ember-engines/lib/engine-addon.js
```diff
@@ -44,6 +44,27 @@ const findRoot = require('./utils/find-root');
...
+  const targetInstanceModule = require('ember-cli/lib/models/per-bundle-addon-cache/target-instance');
+
+  if (targetInstanceModule) {
+    TARGET_INSTANCE_SYMBOL = targetInstanceModule.TARGET_INSTANCE;
+  }
+} catch (e) {
+  // we only want to handle the error when this module isn't found; i.e.,
+  // when a consumer of `ember-engines` is using an old version of `ember-cli`
+  // (less than `ember-cli` 3.28)
+  if (!e || !e.message || !e.message.startsWith('Cannot find module')) {
```
rwjblue, line 63: Let's use `e.code === 'MODULE_NOT_FOUND'` instead

packages/ember-engines/lib/engine-addon.js
```diff
@@ -343,7 +364,13 @@ function buildEngine(options) {
...
 
       hostAddons[addon.name] = addon;
-      queue.push.apply(queue, addon.addons);
+
+      queue.push.apply(
+        queue,
+        TARGET_INSTANCE_SYMBOL && addon[TARGET_INSTANCE_SYMBOL]
+          ? addon[TARGET_INSTANCE_SYMBOL].addons
+          : addon.addons
+      );
```
rwjblue, line 373: I know this is pre-existing, but can you migrate to using spread syntax?
```suggestion
      queue.push(
        TARGET_INSTANCE_SYMBOL && addon[TARGET_INSTANCE_SYMBOL]
          ? ...addon[TARGET_INSTANCE_SYMBOL].addons
          : ...addon.addons
      );
```

rwjblue (changes requested):

rwjblue (approved):

Issue #726, opened 2020-08-13
routeless in-repo-engine blueprint mounts the engine as routed


thousand:
When running `ember g in-repo-engine foo --type routeless`, the blueprint adds a routed mountpoint in `router.js` and doesn't add `{{mount "foo"}}` anywhere.

rwjblue:
Eeck, that's definitely not right! I don't think there is an obvious place to add a `{{mount "foo"}}`,but we definitely should not be adding anything to the `router.js`.

PR #775, opened 2021-05-26
Fix blueprint to routeless in-repo-engine

villander:
it fixes the #726 

The code added is skipping the blueprint to add `mount('foo')` into `router.js` for routeless engines.

cc: @dgeb @rwjblue


rwjblue:
Thanks!

PR #780, opened 2021-06-29
Use @embroider/macros over compatibility helpers

thoov:
Macros is now parallelizable but requires ember-cli-htmlbars v5.3.2 or greater. This is effectively a revert of #752 + the dependency bump.

Related: https://github.com/embroider-build/embroider/issues/882


thoov:
@rwjblue this is now green with your lock file changes

PR #783, opened 2021-07-22
[BUGFIX] Fallback to LinkTo behavior when outside an engine

elwayman02:
Prior to Ember 3.24, LinkToExternal only modifier the `routeKey` when inside an engine; that is, when `owner.mountPoint` exists. With the new class that exists for 3.24+, this behavior regressed. Now, `_namespaceRoute` is overridden in all cases, which means that if `LinkToExternal` ever gets rendered outside of an engine, it will automatically fail since `owner._getExternalRoute` will not find the specified route.

Why this matters: Routing code that may be invoked by an engine does not necessarily live in the same addon as the engine itself. If we extract code that is intended to link to external routes, we still need to be able to test it in those addons, which don't have engines of their own. It's also possible to write reusable code that is designed to be invoked both in an engine and outside of one. The only way to do that is to use LinkToExternal, since the core LinkTo doesn't know that Engines exist. Prior to 3.24, this worked fine thanks to the `owner.mountPoint` check. By adding a similar check here, we ensure that `LinkToExternal` will make a best effort to find the correct route even if it's rendered outside an engine.

### DIFF
--- a/packages/ember-engines/addon/components/link-to-external.js
+++ b/packages/ember-engines/addon/components/link-to-external.js
@@ -11,7 +11,7 @@ if (macroCondition(dependencySatisfies('ember-source', '> 3.24.0-alpha.1'))) {
       const owner = getOwner(this);
       
       if (!owner.mountPoint) {
-        return this._super(...arguments);
+        return super._namespaceRoute(...arguments);
       }
       
       const externalRoute = owner._getExternalRoute(targetRouteName);
--- a/packages/ember-engines/addon/components/link-to-external.js
+++ b/packages/ember-engines/addon/components/link-to-external.js
@@ -9,6 +9,11 @@ if (macroCondition(dependencySatisfies('ember-source', '> 3.24.0-alpha.1'))) {
   LinkToExternal = class LinkToExternal extends LinkComponent {
     _namespaceRoute(targetRouteName) {
       const owner = getOwner(this);
+      
+      if (!owner.mountPoint) {
+        return this._super(...arguments);
+      }
+      
       const externalRoute = owner._getExternalRoute(targetRouteName);
 
       return externalRoute;


villander:
@elwayman02 why you're using `LinkToExternal` outside an engine, the purpose is to use it inside an `engine` to link to external routes.

rwjblue (commented):
Hmm, it actually seems like we should assert here. 🤔 

I can't think of any reason that you would intentionally want to be using `<LinkToExternal />` from **outside** an engine...

elwayman02:
@villander @rwjblue it's not about intentionally using it outside an engine, exactly. It's about writing code that gets used both inside and outside of an engine and needs to work in both places. In particular, as I said above, it's about writing code that is *consumed by an engine* but lives in a different project from the engine itself. We need to be able to test the functionality without having the engine there, since it lives in another project.

I'm not saying it's an ideal design (it certainly wasn't mine), but this is a use-case we have in dozens of places that is blocking us from being able to upgrade to Ember 3.24+. @rwjblue I'm happy to show you the exact way this occurs on Monday; to move away from this design would require a significant amount of inject work unless I literally monkeypatch this component locally in our projects to work around it (which would just be a horrible solution, but is honestly preferable to the alternatives I've thought of if we can't fix this regression).

elwayman02:
Also, the canary build looks like it flaked, can someone re-run it?

villander:
>  it's not about intentionally using it outside an engine, exactly. It's about writing code that gets used both inside and outside of an engine and needs to work in both places. In particular, as I said above, it's about writing code that is consumed by an engine but lives in a different project from the engine itself. We need to be able to test the functionality without having the engine there, since it lives in another project.

I still think it will break the isolation principle of Ember Engines. And the right way to do that is to add it as another use case to be covered on the new Test API on #653

elwayman02:
@villander I agree that this use case should be tested for, but this PR isn't the place to do that (particularly since #653 hasn't even landed). I'm just here trying to fix a behavior regression that absolutely *does* work up until 3.24+.

rwjblue:
After thinking about this a bit more, I think we should land this PR. While I think we all agree that the specific code in question isn't great (a thing should either be external or not), there are other pathways that are just as affected: engine rendering tests. 

Once we move the needle on https://github.com/ember-engines/ember-engines/pull/653, we can revisit this and maybe add a deprecation (because IMO the testing use case is much more important).

PR #789, opened 2021-10-17
fix(build): #405 don't call `included()` twice

### LINKED ISSUE #405: Engines' dependency addons are `include`d more than once

buschtoens:
Fixes #405.

Fixes a bug where the `included()` hook of any addon listed in an engine's `dependencies` would be called twice.

See my comments in #405 for more details.

I have no clue, whether this is actually the correct fix, specifically with regards to `EMBER_ENGINES_ADDON_DEDUPE` introduced via #595. But it's working great for me.

### DIFF
--- a/packages/ember-engines/lib/engine-addon.js
+++ b/packages/ember-engines/lib/engine-addon.js
@@ -605,9 +605,17 @@ function buildEngine(options) {
           };
         }
         originalIncluded.call(this, customHost);
+      } else {
+        // If an `originalIncluded` is present, we assume that it eventually 
+        // delegates to ember-cli's base implementation of `Addon#included()`,
+        // which already calls `included()` for every child addon instance.
+        // As such, we must only call `included()` manually, if there's no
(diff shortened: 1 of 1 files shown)


rwjblue:
@brendenpalmer and I ran into this a bit ago, and it definitely seems like something we ought to try to address. The issue is that it isn't 100% clear that we don't want to call `included` once for each "lazy host". 

@buschtoens - Have you thought of that?

PR #794, opened 2022-01-17
Fix usage of dependencySatisfies with a peerDependency

anehx:
This package uses `dependencySatisfies` of `@embroider/macros` which is
only reliable if a `peerDependency` for the package exists. If not, the
macro checks for a dependency in the `ember-engines` package when a
monorepo structure is used. This behaviour happens since embroider
v0.50.1 which includes an important fix for monorepos:
https://github.com/embroider-build/embroider/pull/1070

### DIFF
--- a/packages/ember-engines/package.json
+++ b/packages/ember-engines/package.json
@@ -86,7 +86,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "@embroider/macros": "^0.43.5",
+    "@embroider/macros": "^1.3.0",
     "amd-name-resolver": "1.3.1",
     "babel-plugin-compact-reexports": "^1.1.0",
     "broccoli-babel-transpiler": "^7.2.0",
@@ -104,6 +104,9 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-version-checker": "^5.1.2",
     "lodash": "^4.17.11"
+  },
+  "peerDependencies": {
+    "ember-source": "^3.12"
   },
   "engines": {
     "node": "10.* || >= 12"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1048,24 +1048,26 @@
     resolve "^1.8.1"
     semver "^7.3.2"
 
-"@embroider/macros@^0.43.5":
-  version "0.43.5"
-  resolved "https://registry.yarnpkg.com/@embroider/macros/-/macros-0.43.5.tgz#f846bb883482436611a58a3512c687d4f9fddfad"
-  integrity sha512-WmLa0T3dyG2XyN5Gr7k1RINDirFzAzh6CRvykRMcuahq1rCrav8ADrWgQzKpPWxdR6DgQCuoyimopJhLbYOpgQ==
+"@embroider/macros@^1.3.0":
+  version "1.3.0"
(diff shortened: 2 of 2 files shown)


anehx:
Similar fix in `ember-render-modifiers` for reference: https://github.com/emberjs/ember-render-modifiers/pull/58

SergeAstapov (approved):

anehx:
How can we get this merged and released? This is blocking a lot of my pipelines..

rwjblue:
I approved the CI run, but it seems that it got cancelled; I'm restarting it

SergeAstapov:
@rwjblue probably it cancels 3.x tests because release/beta get fail too fast at build time because of ember-auto-import v2/webpack miss which is another level of effort

packages/ember-engines/package.json
```diff
@@ -86,7 +86,7 @@
     "ember-addon"
   ],
   "dependencies": {
-    "@embroider/macros": "^0.43.5",
+    "@embroider/macros": "^0.50.2",
```
SergeAstapov, line 89: @anehx `@embroider/macros` is now `1.0.0` so we should be able to do just `"@embroider/macros": "^1.0.0"`

anehx: Done!

SergeAstapov (approved):

anehx:
This is a major issue for me, can I do anything to help getting this released?

packages/ember-engines/package.json
```diff
@@ -105,6 +105,9 @@
     "ember-cli-version-checker": "^5.1.2",
     "lodash": "^4.17.11"
   },
+  "peerDependencies": {
+    "ember-source": "^3.12 || 4"
```
villander, line 109: Why we're adding it here?

anehx: I don't understand the question, could you elaborate? There's a description on why it's needed in the commit / PR body..

villander: I understood the description in the PR body @anehx but I'm trying to understand why you put the `ember-source` in `peerDependencies` instead `devDependencies`. Is it a requirement to have `@embroider/macros` working?

anehx: It's already defined in `devDependencies` which is correct. I added it to `peerDependencies` because the `dependencySatisfies` macro of `@embroider/macros` requires it as peer dependency to be reliable otherwise it won't work properly (e.g. in a monorepo). This is needed because of [this change](https://github.com/embroider-build/embroider/pull/1070) but it's not properly documented yet.

villander: Ok, I got it, but why we're putting 3.12 and 4 here? Since ember-engines are not compatible with ember 4.x yet, I think it's not right

SergeAstapov: @villander looks like you are right, technically we should add v4 only once support is fixed

rwjblue: > technically we should add v4 only once support is fixed
Confirm.

@anehx - Can you update this to be `^3.12.0` for now?

anehx: @rwjblue done, I also updated `@embroider/macros` to the latest version

rwjblue (approved):

PR #797, opened 2022-02-04
Update to ember qunit 5

BobrImperator:
- Hijacks @rwjblue efforts to upgrade to ember-qunit@5 https://github.com/ember-engines/ember-engines/pull/788
- Sets Engine to not autoboot in unit tests
- updates yarn.lock
- Refactors link-to invocation inside `eager-blog`


packages/ember-engines/addon/components/link-to-component.js
```diff
@@ -3,11 +3,11 @@ import { getOwner } from '@ember/application';
 import { computed, get, set } from '@ember/object';
 import { typeOf } from '@ember/utils';
 import { assert, deprecate } from '@ember/debug';
-import { macroCondition, dependencySatisfies } from '@embroider/macros';
+import { macroCondition, getOwnConfig } from '@embroider/macros';
 
 let LinkTo;
 
-if (macroCondition(dependencySatisfies('ember-source', '>= 3.24.1'))) {
+if (macroCondition(getOwnConfig().isEmber324)) {
```
SergeAstapov, line 10: @BobrImperator can we keep using `dependencySatisfies` like in #798?

SergeAstapov:
@BobrImperator do you think we can have this PR focused only on qunit 5?
Ember 4 compatibility can be handled separately in #798 so we can keep PRs focused and simpler to review.

BobrImperator:
Sure @SergeAstapov , I don't think there are any particular changes for Ember 4.
They should be already split, unless you see something I added additionally here.

SergeAstapov:
@BobrImperator there are still some unrelated changes, see https://github.com/ember-engines/ember-engines/pull/797#discussion_r824865553

@BobrImperator I don't think this is needed at all and we should keep using `dependencySatisfies` instead of `ember-cli-version-checker` + `getOwnConfig().isEmber324` in this PR

BobrImperator:
@SergeAstapov true, thanks!
I need to port that over from the other PR

SergeAstapov:
@BobrImperator from my perspective, I would like to keep this PR focused *only* on qunit updates and leave anything related to Ember 4 compat to https://github.com/ember-engines/ember-engines/pull/798, if possible.
Even if CI still not completely happy

SergeAstapov (approved):

runspired (approved):

SergeAstapov:
Closes #788 as it supersedes that effort.

PR #798, opened 2022-02-04
Ember 4 compatibility

BobrImperator:
Built on top of https://github.com/ember-engines/ember-engines/pull/797

- Refactors link-to to LinkTo
- Requires @ember/legacy-built-in-components for ember > 3.24
- Refactors templates to use component/controller properties with `this`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
           - ember-lts-3.20
           - ember-lts-3.24
           - ember-lts-3.28
+          - ember-lts-4.4
           - ember-release
           - ember-beta
           - ember-canary
--- a/packages/ember-engines/addon/-private/engine-ext.js
+++ b/packages/ember-engines/addon/-private/engine-ext.js
@@ -9,7 +9,7 @@ Engine.reopen({
     let registry = this._super(...arguments);
 
     if (!(this instanceof Application)) {
-      if (macroCondition(!dependencySatisfies('ember-source', '>= 3.24.1'))) {
+      if (macroCondition(!dependencySatisfies('ember-source', '>= 3.24.1 || >=4.x'))) {
         const EngineScopedLinkComponent = importSync('ember-engines/components/link-to-component').default;
         registry.register('component:link-to', EngineScopedLinkComponent);
       }
--- a/packages/ember-engines/addon/components/link-to-component.js
+++ b/packages/ember-engines/addon/components/link-to-component.js
@@ -1,11 +1,19 @@
-import LinkComponent from '@ember/routing/link-component';
+import RoutingLinkComponent from '@ember/routing/link-component';
 import { getOwner } from '@ember/application';
 import { computed, get, set } from '@ember/object';
 import { typeOf } from '@ember/utils';
 import { assert, deprecate } from '@ember/debug';
-import { macroCondition, dependencySatisfies } from '@embroider/macros';
+import { macroCondition, dependencySatisfies, importSync } from '@embroider/macros';
 
 let LinkTo;
+let LinkComponent;
+
+if (macroCondition(dependencySatisfies('@ember/legacy-built-in-components', '*'))) {
+  let { LinkComponent: LegacyLinkComponent } = importSync('@ember/legacy-built-in-components');
+  LinkComponent = LegacyLinkComponent;
+} else {
+  LinkComponent = RoutingLinkComponent;
+}
 
 if (macroCondition(dependencySatisfies('ember-source', '>= 3.24.1'))) {
   deprecate(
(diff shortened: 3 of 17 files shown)


packages/ember-engines/addon/components/link-to-component.js
```diff
@@ -1,13 +1,21 @@
-import LinkComponent from '@ember/routing/link-component';
-import { getOwner } from '@ember/application';
+import RoutingLinkComponent from '@ember/routing/link-component';
```
SergeAstapov, line 1: @BobrImperator I may be wrong, but this line would cause build time error with Ember v4. I assume we need to do

```js
LinkComponent = importSync('@ember/routing/link-component');
```

one line 15

packages/ember-engines/addon/components/link-to-component.js
```diff
@@ -1,13 +1,21 @@
...
 import { typeOf } from '@ember/utils';
 import { assert, deprecate } from '@ember/debug';
-import { macroCondition, dependencySatisfies } from '@embroider/macros';
+import { macroCondition, getOwnConfig, importSync, dependencySatisfies } from '@embroider/macros';
 
 let LinkTo;
+let LinkComponent;
 
-if (macroCondition(dependencySatisfies('ember-source', '>= 3.24.1'))) {
+if (macroCondition(dependencySatisfies('@ember/legacy-built-in-components', '*'))) {
```
SergeAstapov, line 11: @BobrImperator I assume we need `@ember/legacy-built-in-components` to be listed in `dependencies` otherwise this line always gonna return `false` in real world app.

As an example you may see #794

BobrImperator: You're right, though for now I assumed that users would need to install it themselves just like in scenarios, but that's not exactly the best idea.

packages/ember-engines/addon/components/link-to-component.js
BobrImperator: This won't work because `@ember` dependencies are transpiled to use global `Ember` in older versions.

Actually the hack is to do `import { LinkTo as RoutingLinkComponent } from '@ember/routing';`
just like in `link-to-external` but for some reason here it's not failing.

This PR is already passing Ember 4

.github/workflows/ci.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           - ember-lts-3.16
           - ember-lts-3.20
           - ember-lts-3.24
+          - ember-lts-3.28
```
villander, line 48: Where's the Ember 4 here in the package.json?

packages/ember-engines/index.js
```diff
@@ -20,4 +39,15 @@ module.exports = ManifestGenerator.extend({
     manifest.vendorFiles.push('assets/node-asset-manifest.js');
     return manifest;
   },
+
+  _findHost() {
+       let current = this;
```
villander, line 44: Please fix the indentation

packages/ember-engines/index.js
```diff
@@ -20,4 +39,15 @@ module.exports = ManifestGenerator.extend({
     manifest.vendorFiles.push('assets/node-asset-manifest.js');
     return manifest;
   },
+
+  _findHost() {
```
villander, line 43: It has not been used anywhere. Why we need it?

packages/ember-engines/addon/components/link-to-component.js
```diff
@@ -1,13 +1,21 @@
...
+  let { LinkComponent: LegacyLinkComponent } = importSync('@ember/legacy-built-in-components');
```
villander, line 12: Thank you for doing that @BobrImperator. But I prefer the solution made by @SergeAstapov [here](https://github.com/ember-engines/ember-engines/pull/796) it's more composable 😄

.github/workflows/ci.yml
BobrImperator: ember-canary, release and beta channels are the ember 4.
I'm not sure whats the usual practice now for testing ember 4 compatibility, but preferably we should test only against LTS versions when there is no v4 LTS yet?

packages/ember-engines/index.js
BobrImperator: We don't ✂️🙏

packages/ember-engines/addon/components/link-to-component.js
BobrImperator: I'd prefer it too, but it's not working yet if I understand the description correctly.

We can rebase after/before it lands

.github/workflows/ci.yml
rwjblue: Ya, I generally agree. But since Ember does publish alphas from master (and master is 4.4.x), we can add `ember-lts-4.4` and use `^4.4.0-alpha.1` in ember-try config.

rwjblue (changes requested):

.github/workflows/ci.yml
BobrImperator: Thanks, feedback applied.

packages/ember-engines/addon/-private/engine-ext.js
```diff
@@ -9,7 +9,7 @@ Engine.reopen({
     let registry = this._super(...arguments);
 
     if (!(this instanceof Application)) {
-      if (macroCondition(!dependencySatisfies('ember-source', '>= 3.24.1'))) {
+      if (macroCondition(!getOwnConfig().isNewerThanEmber324)) {
```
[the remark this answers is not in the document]
BobrImperator: The reason for this was that the macros were misbehaving, for some reason the check was constant between ember-source versions which was breaking.
So I thought about doing whatever I did, but ultimately, I just changed the match string to `^3.24.0 || ^4.x` and now it works correctly.

packages/ember-engines/tests/acceptance/routeable-engine-demo-test.js
```diff
@@ -14,13 +14,13 @@ module('Acceptance | routable engine demo', function (hooks) {
   setupApplicationTest(hooks);
 
   hooks.beforeEach(function () {
-    originalExceptionHandler = Ember.Test.adapter.exception;
+    originalExceptionHandler = TestAdapter.exception;
```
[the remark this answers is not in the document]
BobrImperator: Yes, you're right :)
Removed.

packages/ember-engines/addon/-private/engine-ext.js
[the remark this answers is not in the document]
SergeAstapov: @BobrImperator I think you hit an issue that was fixed by #794. Could you please check if that's the case and we can keep using `dependencySatisfies`?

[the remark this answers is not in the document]
BobrImperator: @SergeAstapov It should be resolved already, the code no longer uses custom config

rwjblue (approved):

PR #808, opened 2022-07-28
Move broccoli-test-helper to devDependencies

ginomiglio:
This file is only used in `node-tests/` so it doesn't need to be installed by consuming apps/addons. (I noticed this because it was causing my app to include a very old version of [ember-cli/watch-detector](https://github.com/ember-cli/watch-detector).)


SergeAstapov:
@ginomiglio thanks for contribution and sorry for delay. do you mind to refresh your branch as now there is a conflict?

ginomiglio:
@SergeAstapov done!

SergeAstapov:
Thank you @ginomiglio!

PR #810, opened 2022-08-03
Refactor LinkToExternal to use LinkTo directly

BobrImperator:
blocked by https://github.com/emberjs/ember.js/pull/20151

- removes dependency on @ember/legacy-built-in-components introduced here https://github.com/ember-engines/ember-engines/pull/798
- Removes tests invoking link-to-external with positional args
  - this is possibly a breaking change, although in Ember v4 LinkTo already doesn't handle that.

### DIFF
--- a/packages/ember-engines/addon/templates/link-to-external-template.hbs
+++ b/packages/ember-engines/addon/templates/link-to-external-template.hbs
@@ -1,0 +1,2 @@
+<LinkTo @route={{this.route}} @isExternal={{true}} ...attributes>{{yield}}</LinkTo> 
+
--- a/packages/ember-engines/addon/components/link-to-component.js
+++ b/packages/ember-engines/addon/components/link-to-component.js
@@ -5,17 +5,9 @@ import { getOwner } from '@ember/application';
 import { computed, get, set } from '@ember/object';
 import { typeOf } from '@ember/utils';
 import { assert, deprecate } from '@ember/debug';
-import { macroCondition, dependencySatisfies, importSync } from '@embroider/macros';
+import { macroCondition, dependencySatisfies } from '@embroider/macros';
 
 let LinkTo;
-let LinkComponent;
-
-if (macroCondition(dependencySatisfies('@ember/legacy-built-in-components', '*'))) {
-  let { LinkComponent: LegacyLinkComponent } = importSync('@ember/legacy-built-in-components');
-  LinkComponent = LegacyLinkComponent;
-} else {
-  LinkComponent = RoutingLinkComponent;
-}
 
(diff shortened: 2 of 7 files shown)


packages/ember-engines/addon/templates/link-to-external-template.hbs
```diff
@@ -0,0 +1,2 @@
+<LinkTo @route={{this.route}} @isExternal={{true}} ...attributes>{{yield}}</LinkTo>
```
SergeAstapov, line 1: @BobrImperator personally I think this solves most but not all the ways `<LinkTo />` (and subsequently `<LinkToExternal />`) components can be used. e.g. it does not support providing query params or model for the external route.

we believe `<LinkToExternal />` should have exact same interface as `<LinkTo />` and re-implementing that interface in `ember-engines` is brittle and would lead to a lot of duplication.

For that reason, RFC https://github.com/emberjs/rfcs/pull/819 was opened to solve this problem in a way that would not require duplication nor use of intimate API.

cc @villander @runspired

villander: I agree with @SergeAstapov given the `LinkToExternal` is part of the first Ember Engines RFC we need to approve the new one to accept this implementation. That being said the RFC is the best way to build things right not just quickly. 

I appreciate your contribution @BobrImperator and would love to get you involved in the RFC https://github.com/emberjs/rfcs/pull/819

BobrImperator: Thanks for the feedback.
I forgot about all other args... would need to proxy them.

If you don't see it as feasible then it's fine with me. My main concern however is getting rid of the legacy components hack that I shouldn't have added in the first place. But had no better idea at the time.

Although, fixing it this way should allow people to more easily migrate to ember 4, in fact the reason I took another look at this was that embroider has disabled it's tests against engines because there was no support at the time while technically there's support finally but it's really hard to use at the moment.

https://github.com/embroider-build/embroider/pull/1123

SergeAstapov: @BobrImperator as https://github.com/emberjs/rfcs/pull/819 takes some time to mature and land, using `@ember/legacy-built-in-components` as temp solution is totally fine as this allows community to use `ember-engines` with Ember.js 4+ until long-term solution lands.

there is nothing necessarily "bad" about `@ember/legacy-built-in-components`, it's just an extra code that needs to be shipped, but will get rid of it once https://github.com/emberjs/rfcs/pull/819 lands and implemented.

void-mAlex:
superseeded by #855

PR #812, opened 2022-08-10
Add deprecation warning to `engineResolverFor`

villander:
It adds a deprecation warning to `engineResolverFor` test helper given the new test API released at #653 

cc // @SergeAstapov @runspired

### DIFF
--- a/packages/ember-engines/addon-test-support/engine-resolver-for.js
+++ b/packages/ember-engines/addon-test-support/engine-resolver-for.js
@@ -1,5 +1,6 @@
 /* global require */
 import EmberResolver from 'ember-resolver';
+import { deprecate } from '@ember/debug';
 
 /**
  * Gets the resolver class used by an Engine and creates an instance to be used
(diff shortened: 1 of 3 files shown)


packages/ember-engines/addon-test-support/engine-resolver-for.js
```diff
@@ -29,6 +30,19 @@ export default function engineResolverFor(
...
+  deprecate(
+    "Use of `engineResolverFor` has been deprecated. Instead use `setupEngine(hooks, 'engine-name')` imported from `ember-engines/test-support` to load the engine you need.",
+    false,
+    {
+      id: 'ember-engines.addon-test-support.engine-resolver-for',
+      for: 'ember-engines',
+      until: '1.0.0',
+      since: {
+        enabled: '0.8.23',
+        available: '0.8.23',
```
SergeAstapov, line 42: @villander these actually needs to be `0.8.24` because current version is `0.8.23` and this will be included in the *next* release

PR #855, opened 2023-12-19
enable ember v5 compatibility and add embroider to try scenarios

void-mAlex:
see [Upgrade ](https://github.com/ember-engines/ember-engines/blob/2b0924edbcb4fd91672e73eafc00d2dc9d246c20/packages/ember-engines/UPGRADE.md) guide for details


packages/ember-engines/addon/components/link-to-external.js
```diff
@@ -1,53 +1,933 @@
...
+        }
+        if (!isSelf) {
+            return false;
+        }
+        let { _route: routeName, _models: models, _query: queryParams, replace: shouldReplace, } = this;
+        let payload = {
+            queryParams,
+            routeName,
+        };
+        // flaggedInstrument(
```
SergeAstapov, line 708: @void-mAlex if we go this route, how could we ensure this component gets updated and we do not fall out of sync with `ember-source`?

void-mAlex: this is not from ember source, it's the legacy-component one that will never ever be updated again

SergeAstapov: @void-mAlex if I remember correctly, we end up using `@ember/legacy-built-in-components` because we could import from and extend the class.
And we wanted to use `LinkTo` provided by `ember-source` but it's not importable any more.

If we go this route, why don't we copy-paste `LinkTo` from `ember-source`? am I missing something?

packages/ember-engines/package.json
```diff
@@ -103,22 +102,17 @@
...
-  },
-  "peerDependenciesMeta": {
-    "@ember/legacy-built-in-components": {
-      "optional": true
-    }
+    "ember-source": "^3.24.1 || 4 || >=5"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
+    "node": "16.* || >= 18"
```
SergeAstapov, line 115: @void-mAlex I think we should do this in separate PR and probably go 18+, wdyt?

packages/ember-engines/package.json
```diff
@@ -103,22 +102,17 @@
     "calculate-cache-key-for-tree": "^2.0.0",
     "ember-asset-loader": "^1.0.0",
     "ember-cli-babel": "^7.18.0",
+    "ember-cli-htmlbars": "^5.3.2",
```
SergeAstapov, line 105: maybe go v6?

packages/ember-engines/addon/components/link-to-external.js
void-mAlex: this way I can guarantee that the only 'breaking' change you really need is the re-export
I'm not sure exactly everything the new link to drops that the legacy one was handling for engines specifically, this way it's the least likely to cause changes in behaviour

trying to tide people over till there is a better router as I don't want to be changing this more than necessary myself

packages/ember-engines/package.json
void-mAlex: happy to do 18+, but not sure about value of doing it as a separate pr, they still need to go in together
happy to break them up but more work imo when you really end up doing two separate breaking changes

packages/ember-engines/package.json
void-mAlex: waiting on ci to see it green then can bump this to 6 and node to 18 in one go

packages/ember-engines/addon/components/link-to-external.js
SergeAstapov: @void-mAlex fair point!

void-mAlex:
@SergeAstapov this adds a few more breaking changes like getting in line with rfc 674
this version would mean that people could go to ember-source v5 if they can change over to using the router service and don't use query param only transitions which don't think it would be too hard to fix suport for in ember-engines-router-service

@villander thoughts on all of this?

void-mAlex:
@SergeAstapov  the easy fix to get ci to run was to bring back node 16 as it was running up against the crypto package used by terser-webpack-plugin
which as far as I can tell only affects the testing side of things and not the runtime of the package.
I don't think we should add any more to this pr in terms of updating dependencies, but open to hearing  other people's thoughts on it
this would make for a good compromise on path to v5 for ember

void-mAlex:
@SergeAstapov any idea why the try scenario might be failling now? they are all fine locally
I'm just about giving up on those try scenarios as something is not working with the scenarios themselves rather than the rest of the code
and this would now be ready for a review if you're up for it

void-mAlex:
all green 🎉  
@SergeAstapov

.github/workflows/ci.yml
```diff
@@ -58,7 +59,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 20.x
```
void-mAlex, line 62: got these running against node 20 which is not required, but gives us a bit more coverage in terns of testing
node 18+ has issues for the node tests in the non ember-try run which seem to be test infrastructure related only and would not affect users

packages/ember-engines/package.json
villander: I agree with @void-mAlex here, we don't need a separated PR when we can hit two birds with one stone ; )

villander:
@void-mAlex let me know if I understood it correctly: Can't we use the query params with these changes? If yes I think that's a big regression 

https://github.com/ember-engines/ember-engines/pull/855#issuecomment-1865261182

packages/ember-engines/package.json
```diff
@@ -50,38 +54,41 @@
...
     "ember-cli-app-version": "^3.2.0",
     "ember-cli-dependency-checker": "^3.2.0",
-    "ember-cli-htmlbars": "^5.3.2",
-    "ember-cli-inject-live-reload": "^2.0.2",
+    "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-sri": "^2.1.1",
-    "ember-cli-uglify": "^3.0.0",
+    "ember-cli-terser": "^4.0.2",
     "ember-disable-prototype-extensions": "^1.1.3",
+    "ember-engines-router-service": "^0.5.0",
```
villander, line 65: ❤️

void-mAlex:
> @void-mAlex let me know if I understood it correctly: Can't we use the query params with these changes? If yes I think that's a big regression
> 
> [#855 (comment)](https://github.com/ember-engines/ember-engines/pull/855#issuecomment-1865261182)

I'm glad I wrote this up initially 
see upgrade guide explains what that means 
https://github.com/ember-engines/ember-engines/pull/855/files#diff-8a5dff0178144a6ddc5d4c44d16a86bbfa2ec3e8da9cb204abad70b8b39712f0R22

void-mAlex:
so to clarify for anyone not following the links, you can still use query params, there is a specific subcase where you transtition without a route name that appears to be an open bug at the time of writing

packages/ember-engines/addon/components/link-to-external.js
```diff
@@ -1,53 +1,941 @@
...
-  LinkToExternal = class LinkToExternal extends LinkComponent {
-    _namespaceRoute(targetRouteName) {
-      const owner = getOwner(this);
-     
-      if (!owner.mountPoint) {
-        return super._namespaceRoute(...arguments);
-      }
-      
-      const externalRoute = owner._getExternalRoute(targetRouteName);
+/**
```
villander, line 49: @void-mAlex is it a copy/paste of original LinkTo?

villander:
> so to clarify for anyone not following the links, you can still use query params, there is a specific subcase where you transtition without a route name that appears to be an open bug at the time of writing

Ok, we can fix it later in the `ember-engines-router-service` addon. Thanks for the hard work here!

villander (approved):
Thanks for the hard work!

villander:
@SergeAstapov that sounds reasonable for moving forward and make engines compatible with v5, if you don't have any blocks please let's move forward and release it 🔥

packages/ember-engines/addon/components/link-to-external.js
void-mAlex: short answer is yes, long one is it's not byte for byte for reasons of deprecation removals

packages/ember-engines/UPGRADE.md
```diff
@@ -20,5 +20,7 @@
     > migrate to using router service from ember-engines-router-service
 
 5. drop support for query param only router service transitions until it's resolved upstream https://github.com/villander/ember-engines-router-service/issues/74
+    > example for unsupported syntax ❌ `this.router.transitionTo({ queryParams: { page: 1 }});` without passing a route name
```
villander, line 23: Great documentation to improve ergonomic usage. Thank you @void-mAlex

packages/ember-engines/addon/components/link-to-external.hbs
```diff
@@ -0,0 +1,5 @@
+{{~#if (has-block)~}}
+  {{yield}}
+{{~else~}}
+  {{this.linkTitle}}
+{{~/if~}}
```
villander, line 5: @void-mAlex can we change this syntax? It's not what we're used to seeing in the ember guides

void-mAlex: it's exactly what https://github.com/emberjs/ember-legacy-built-in-components/blob/095fa3eba7cddf0716ec89fd4bb36087ebe2662f/addon/components/link-to.hbs
the original had which I kept intact
I would advocate for minimal changes with this as while it looks funky all it does is remove whitespace, and that could be a change in behaviour for consumers
so unless you feel really strongly about this I wouldn't touch it

packages/ember-engines/UPGRADE.md
void-mAlex: @villander is that issue something you'd actively try to fix? don't think it's something widely used, but obviously people do use it since it was reported.

packages/ember-engines/addon/-private/engine-ext.js
```diff
@@ -1,16 +0,0 @@
...
-import Engine from '@ember/engine';
-
-import ExternalLinkComponent from '../components/link-to-external';
-
-Engine.reopen({
-  buildRegistry() {
-    let registry = this._super(...arguments);
-
-    if (!(this instanceof Application)) {
-      registry.register('component:link-to-external', ExternalLinkComponent);
```
SergeAstapov, line 11: @void-mAlex any specific reason we can't keep this and do not require engine users to re-export `link-to-external` themselves?

packages/ember-engines/package.json
```diff
@@ -102,20 +108,16 @@
...
   },
   "peerDependencies": {
-    "ember-source": "^3.24.1 || 4 || 5",
-    "@ember/legacy-built-in-components": "*"
-  },
-  "peerDependenciesMeta": {
-    "@ember/legacy-built-in-components": {
-      "optional": true
-    }
+    "ember-engines-router-service": ">=0.5.0",
```
SergeAstapov, line 119: @void-mAlex if I'm reading `packages/ember-engines/addon/components/link-to-external.js` correctly, it does not depend on `ember-engines-router-service`.

am I get it right that only reason for this peer dependency - we delete `transitionToExternal`/`replaceWithExternal` route methods and `transitionToExternalRoute` controller methods because they remove in v5 as explained in https://deprecations.emberjs.com/v3.x#toc_routing-transition-methods?

If so, I think this needs to be extracted to separate PR as it's own deprecation. I can do that

packages/ember-engines/UPGRADE.md
SergeAstapov: @villander did you had a chance to look into it? IMO would be great to fix that and at least https://github.com/villander/ember-engines-router-service/pull/95 before we recommend ppl using that addon OOTB

packages/ember-engines/package.json
SergeAstapov: also, why don't we make `ember-engines-router-service` a dependency of `ember-engines` so ppl don't need to install it themselves?

void-mAlex: > also, why don't we make `ember-engines-router-service` a dependency of `ember-engines` so ppl don't need to install it themselves?

simply because it logically doesn't have enough enough knowledge to handle external routes if it was scoped to only the engine
it's an app concern around it's routes and we shouldn't force it into peoples' builds that don't use routable engines

ps I'll try and answer the first comment a bit later

SergeAstapov: > we shouldn't force it into peoples' builds that don't use routable engines

sounds like this should be optional peer dependency then?

void-mAlex: > @void-mAlex if I'm reading `packages/ember-engines/addon/components/link-to-external.js` correctly, it does not depend on `ember-engines-router-service`.

correct

> am I get it right that only reason for this peer dependency - we delete `transitionToExternal`/`replaceWithExternal` route methods and `transitionToExternalRoute` controller methods because they remove in v5 as explained in https://deprecations.emberjs.com/v3.x#toc_routing-transition-methods?

also correct and it should be optional as you previously stated

> If so, I think this needs to be extracted to separate PR as it's own deprecation. I can do that

ty

villander (approved):
Amazing! Thank you for the hard work here.

patches/@embroider__router@2.1.7.patch
```diff
@@ -0,0 +1,35 @@
+diff --git a/dist/index.js b/dist/index.js
+index 2df582981f388c3cc8af7db9a2fb288710c4727c..2c4757d2dfd602ae2725a0cba01780e466d7565a 100644
+--- a/dist/index.js
```
villander, line 3: @void-mAlex why do we need this?

packages/ember-engines/tests/test-helper.js
```diff
@@ -14,7 +28,14 @@ setApplication(Application.create(config.APP));
 setup(QUnit.assert);
 setupSinon();
 
-preloadAssets(manifest).then(start); // This ensures all engine resources are loaded before the tests
+if (macroCondition(!dependencySatisfies('@embroider/core', '*'))) {
+  preloadAssets(manifest).then(start); // This ensures all engine resources are loaded before the tests
```
villander, line 32: we shouldn't preload assets anymore. Could you please make the tests more real?

patches/@embroider__router@2.1.7.patch
void-mAlex: we don't need it
it's was to prove work related to https://github.com/embroider-build/embroider/pull/1945
got that fixed upstream and will upgrade to that version and drop the patch

packages/ember-engines/tests/test-helper.js
void-mAlex: how do you propose tests be made more real?
deleting these lines makes the tests more real but engines code doesn't work to on the fly load the necessary code where embroider does exactly what it should
that's why for the tests  where it's still needed the assets get preloaded and where it's not we're making the tests better by not preloading them at all

packages/ember-engines/UPGRADE.md
```diff
@@ -0,0 +1,24 @@
+# ember-engines
+
+## Breaking changes and how to resolve them
+
+### 0.10.0
```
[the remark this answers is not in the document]
void-mAlex: keen on getting this version out and we can iterate on documentation

packages/ember-engines/addon/-private/engine-ext.js
void-mAlex: people shouldn't need to re-export it themselves as it's provided by an app re-export within ember-engines itself
this just makes the system a bit saner and not assume things about the registry existing and being the thing that is used to resolve components (gets replaced under embrodier staticComponents with compile time lookups)

packages/ember-engines/README.md
```diff
@@ -19,6 +19,9 @@ perspective.
 ember install ember-engines
 ```
 
+## Upgrade guide
+
+* [0.10.0 -> 0.11.0](UPGRADE.md#0110)
```
SergeAstapov, line 24: @void-mAlex there is `Migrating` section in https://ember-engines.com/docs/ so I believe we should keep documentation consolidated

packages/ember-engines/package.json
SergeAstapov: @void-mAlex so `@ember/legacy-built-in-components` is still listed in `peerDependenciesMeta`, I think you want to replace it with `ember-engines-router-service` as we want `ember-engines-router-service` be optional and it's not a requirement

packages/ember-engines/tests/acceptance/lazy-routeable-engine-test.js
```diff
@@ -81,15 +81,15 @@ module('Acceptance | lazy routable engine', function (hooks) {
     );
   }
 
-  test('it should pause to load JS and CSS assets on deep link into a lazy Engine', async function (assert) {
+  test('@classiconly it should pause to load JS and CSS assets on deep link into a lazy Engine', async function (assert) {
```
SergeAstapov, line 84: @void-mAlex should you add `@classiconly` tag to the module (now it says `Acceptance | lazy routable engine`) so you don't add it to each individual test?

packages/ember-engines/tests/dummy/app/router.js
```diff
@@ -1,4 +1,4 @@
-import EmberRouter from '@ember/routing/router';
+import EmberRouter from '@embroider/router';
```
SergeAstapov, line 1: @void-mAlex is this *required* to work with `ember-source` v5? Or, is it required only for Embroider scenarios?

packages/ember-engines/tests/dummy/config/ember-try.js
```diff
@@ -47,6 +45,24 @@ module.exports = async function () {
         npm: {
           devDependencies: {
             'ember-source': '~5.4.0',
+            '@ember/string': '*',
+            'ember-engines-router-service': '*',
```
SergeAstapov, line 49: @void-mAlex this is strange it's added only here and not added to other scenarios.

Also I thought you would want `ember-engines-router-service` in `devDependencies` of `ember-engines` package - otherwise how does it get pulled into build at all?

packages/ember-engines/tests/test-helper.js
SergeAstapov: @void-mAlex so for this to work, we would need to provide re-exports in the dummy app for each and every service/component/helper/modifier/route/controller/etc.?

sounds like `setupEngine` test helper won't work moving forward

packages/ember-blog/.eslintrc
```diff
@@ -1,9 +0,0 @@
-{
```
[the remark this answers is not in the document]
void-mAlex: this one I had to get rid of as it was incorrect/not working after the eslint upgrades. simply producing errors in the console, the higher level eslint still applies and I don't see why we need independent lint configs for all separate projects(especially when they don't work)

packages/ember-engines/tests/test-helper.js
void-mAlex: setupEngine needs to go the way of the dodo
and to answer your question, quite possibly for testing the components of the engine itself
though you have to pretty much import explicitly all your components in tests in gts format ... so this is the in between step to getting there

packages/ember-blog/addon/components/link-to-external.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ember-engines/components/link-to-external';
```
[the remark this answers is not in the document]
void-mAlex: need the re-export as ember-engines is not a dependency of ember-blog
the whole setup in the monorepo is plenty wrong as the 'test engines' which I can barely call them aren't even set up as an engine would normally be (currently they are not even considered ember-cli projects)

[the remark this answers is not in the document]
void-mAlex: what is the downside of apps having access to this component?

packages/ember-engines/README.md
void-mAlex: happy to consolidate there as well, I don't see how this extra doc hurts

.gitignore
```diff
@@ -23,3 +23,4 @@ testem.log
 .node_modules.ember-try/
 bower.json.ember-try
 package.json.ember-try
+packages/ember-engines/.eslintcache
```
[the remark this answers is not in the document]
void-mAlex: think I asked vs code to ignore it for me and that's where it put it, don't mind either way

packages/ember-engines/tests/acceptance/lazy-routeable-engine-test.js
void-mAlex: there are three tests in the module that don't have the identifier

packages/ember-engines/tests/dummy/app/router.js
void-mAlex: embroider, but I'm not separating the work, there is no point in doing so, having 2 prs doesn't change that this all needs to happen and it's a pain to pull it apart
happy to rename the pr to something more encompassing

packages/ember-engines/package.json
void-mAlex: nice catch, ty

packages/ember-engines/tests/acceptance/routeable-engine-demo-test.js
```diff
@@ -291,9 +291,6 @@ module('Acceptance | routable engine demo', function (hooks) {
     assert
       .dom('.routable-post-transition-to-home-button')
       .hasClass('transitioned-to-external');
-    assert.deprecationsInclude(
```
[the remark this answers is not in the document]
void-mAlex: this is actually required for ember source v5
as it required the deletion of 
packages/ember-engines/addon/-private/route-ext.js
and other ext files as the things they were extending are gone

packages/ember-engines/tests/dummy/app/router.js
void-mAlex: last time we split things apart ember v5 support snuck out early by way of peer dep relaxing wihtout actual support so if people tried it it would fail on them (unless it's a new project)

SergeAstapov (approved):

SergeAstapov (approved):

PR #864, opened 2024-03-09
Migrate internal tests to new testing API

villander:
It refactors the current tests for the new **Ember Engines Testing API** #653 

The purpose of #653 was to create an API for ember engines that gives our audience the same approach to test engines as it has for normal ember apps and add-ons without `preloadAssests` and `engineResolveFor`. I discovered an issue when we tried to import some files not found by `this.engine.lookup()` for example `import Initializer from 'ember-blog/initializers/ember-blog-initializer';` at this [line](https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/tests/acceptance/routeable-engine-demo-test.js#L3)

I've tried the usage of `importSync` from embroider into the test suite to load these files on-demand but it didn't work for initializers, given the engine is [booted during the `setupEngine()` calling](https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/addon-test-support/index.js#L66). That being said we still need some work to make the new API consistent with the [Ember guides](https://guides.emberjs.com/release/testing/) where we can import any files from the app/addon into the test files. 

image

cc @SergeAstapov @void-mAlex @chriskrycho


void-mAlex (changes requested):
preloadAssets and setupEngine are actually both very bad as they rely on asset manifest which is not compatible with the future of ember and es modules in general
instead of this we really need to deprecate both approaches immediately

villander:
> preloadAssets and setupEngine are actually both very bad as they rely on asset manifest which is not compatible with the future of ember and es modules in general
instead of this we really need to deprecate both approaches immediately

I kindly disagree because It's not true, `setupEngine` is way better than `preloadAssets` and closer to the ember-engines in real life, loading engines on demand. You're worrying about the implementation instead of thinking about our users and how painful it's been for many years. I'd love to discuss it with you, @SergeAstapov, and anyone else interested, I'm pretty sure you're missing some gaps about what Ember Engines in the Ember ecosystem @void-mAlex. 

I think everyone here knows the `ember-asset-loader` will be deprecated in favor of embroider (I might be one of the person that most want to see it), but erasing the whole RFC as you're doing is not the best thing to do here, please remember we have an RFC and we still rely on it until a new ember engines/micro frontend comes up. 🙏

void-mAlex:
I'm blocking this pr as it would fail CI and it's the current changes are the wrong thing to do even if you want to move to the new testing api
steps would be 
1. make the engines have a proper structure as one would expect to see an engine following the guides to set one up
2. move the preload assets to setupEngine with the macros in place similar to what the current test-helper does to preload assets so we can have clean CI
3. implement and switch to a setupEngine provided by embroider to make testing engines possible under those scenarios as well (keep macro config for backwards compatibility tests)

that is the direction this pr needs to take, but most likely need to land step one in separate smaller prs that are more manageable

villander:
@void-mAlex thanks for the response. Bringing clarity to this conversation the `setupEngine` is an official helper merged and approved by the core team into this addon, so the tests were passing and we were using this official test helper to modernize the code. That being said I don't see any reason to block this PR, what you are saying is 100% out of the scope of this PR proposal.

void-mAlex:
@villander resolve the conflict and you will see what I'm talking about with failing tests
second it's very related to this pr as those tests are in the wrong place. even if you stretch the intent of the tests engines have never been meant to work in a v1 addon which is where they are working from now. all that we are testing now is that that engines 'work' in a place where they are not supposed to ever be used. Add on top of that, that the things being tested are not even able to be considered ember projects let alone engines there is no reason to move forward with something that doesn't address the underlying problem and just makes more code that needs to be cleaned up.

villander:
@void-mAlex we still need to make the tests modernized and don't load everything once, we need the ability of test engines as they perform in real life. You're not proposing any changes here to keep the Testing user story getting better.

SergeAstapov:
@villander @void-mAlex I think there is miscommunication in this tread.

Me and @void-mAlex had a discussion about this and using `setupEngine()` test helper for is not a problem per se.
The problem is that `setupEngine()` under the hood uses asset loader which is not supported under Embroider.

Once we figure out how we could load and instantiate engine using Embroider router or Embroider provided tools under the hood - we should be good.

With above, public API of `setupEngine()` should not change even under Embroider:

```js
import MyComponent from 'my-engine/components/my-component';
import { render } from '@ember/test-helpers';

module('Rendering | MyComponent', function (hooks) {
  setupRenderingTest(hooks);
  setupEngine(hooks, 'my-engine');

  test('it renders', function () {
    await render(<template><MyComponent /></template>, { owner: this.engine });
    assert.dom('[data-test-my-component]').exists();
  });
});
```

the tricky part - how to create an instance of engine under Embroider. That's something @void-mAlex had some ideas for.

I agree with @void-mAlex in that this PR "as is" can't go in as it would break CI and we need to do some maintenance in this repo to properly setup engines as addons and test components provided by those engines in the engine's dumm app (rather than doing it in `ember-engines` package which is wrong)

villander:
@SergeAstapov can we have another discussion about the `setupEngine` with the three of us, I have a different option about this direction, given we're still relying on `ember-asset-loader` and the CI is green now with the engines being built upon this addon I don't see any problem in fixing this PR to make it green so we can handle with `ember-asset-loader` later and hit two birds with one stone, also I can lead this initiative in replacing the ember-asset-loader by Embroider router/loading given the embroider router is not enough because we have routeless engines. 

> the tricky part - how to create an instance of engine under Embroider.

I also have some ideas, how we can discuss it as a team? 

> I agree with @void-mAlex in that this PR "as is" can't go in as it would break CI 

can we merge it if I get the CI green @SergeAstapov?

SergeAstapov:
> can we merge it if I get the CI green?

@villander tbh I don't think is possible under Embroider. I would be glad to be wrong, so you could make it green - I would be happy to review proposed changes!

villander:
@SergeAstapov we don't need to make it work under embroider right away - we just need to do the same as we did for #855 so we can make the tests real satisfying the original RFC which means the `@classiconly` so we can handle embroider later. 

```ts
if (macroCondition(!dependencySatisfies('@embroider/core', '*')) {
  // handle with ember asset loader stuff
}
```

villander:
@SergeAstapov can we merge this PR? It will be a big game changer to the community

void-mAlex:
@villander resolve the conflicts and make it green without skipping embroider tests and I will help land this

void-mAlex:
as I said before I'll help land this pr if all of CI is green

villander:
@void-mAlex, all CI is green. Would you mind reviewing and merging?

SergeAstapov (approved):

packages/ember-engines/tests/acceptance/routeable-engine-demo-test.js
```diff
@@ -360,15 +363,33 @@ module('Acceptance | routable engine demo', function (hooks) {
...
 
     assert.ok(stub.calledOnce, 'Initializer ran once');
   });
 
   test('instance initializers run within engine', async function (assert) {
-    let stub = sinon.stub(InstanceInitializer, 'initialize');
+    if (macroCondition(!dependencySatisfies('@embroider/core', '*'))) {
+      let assetLoader = this.owner.lookup('service:asset-loader');
+      await assetLoader.loadBundle('ember-blog');
+    }
```
void-mAlex, line 386: not sure what this change accomplishes 
test seems to pass well enough without any of it

villander: If you remove the **if** here, it will break on CI. Could you please give a try?

void-mAlex (approved):

PR #867, opened 2024-03-24
Switch package manager to pnpm

void-mAlex:
as the title says this switches the package manager to latest pnpm version as of pr creation 8.15.5
it comments out the known ember v5 try scenarios to get a green pr with the intent this lands before #855 that fixes newer ember compatibility

this pr is a pre-req for adding release-plan so we eliminate the need to have npm access granted as it can be set up as a github secret and have an action do the publish for us which is also more secure and makespermission management easier


SergeAstapov:
Thank you @void-mAlex!

villander:
great job @void-mAlex thanks for pushing it forward

PR #868, opened 2024-03-25
Convert codebase to use native classes

SergeAstapov:
Extract from #855 to keep it focused

### DIFF
--- a/packages/ember-blog/addon/routes/post.js
+++ b/packages/ember-blog/addon/routes/post.js
@@ -1,29 +1,30 @@
 import Route from '@ember/routing/route';
+import { action } from '@ember/object';
 
-export default Route.extend({
+export default class PostRoute extends Route {
   model(params) {
     return {
       user: this.modelFor('application'),
       id: params.id,
-      title: `Post ${params.id}`,
+      title: `Post ${params.id}`
     };
-  },
+  }
 
-  actions: {
-    goToChineseVersion() {
-      this.transitionTo({ queryParams: { lang: 'Chinese' } });
-    },
-    transitionToHome() {
-      this.transitionToExternal('home').then(() => {
-        var postController = this.controllerFor(this.routeName);
-        postController.set('transitionedToExternal', true);
-      });
-    },
-    replaceWithHome() {
-      this.replaceWithExternal('home').then(() => {
-        var postController = this.controllerFor(this.routeName);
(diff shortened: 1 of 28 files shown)


packages/ember-blog/addon/routes/post.js
```diff
@@ -1,29 +1,30 @@
...
-    replaceWithHome() {
-      this.replaceWithExternal('home').then(() => {
-        var postController = this.controllerFor(this.routeName);
-        postController.set('replacedWithExternal', true);
-      });
-    },
-  },
-});
+  @action goToChineseVersion() {
+    this.transitionTo({ queryParams: { lang: 'Chinese' } });
```
void-mAlex, line 14: converting this in isolation of the 855 pr change? if yes then that pr will need need to recharge this line to make tests work with e-engines-router-service due to a known bug with qp only transitions

villander:
@SergeAstapov I don't think we need to fix it first before merging the #855 could you please clarify what you're trying to achieve here enriching the PR description?

packages/ember-blog/addon/routes/post.js
SergeAstapov: @void-mAlex sorry I think I missed your point. This PR focuses solely on syntax change, without any functional changes.

Once we land this, we we would simply need to rebase #855, do all the necessary functional tweaks without noise of syntax conversion and land it with sole focus on v5 support via using e-engines-router-service

villander: Why do we need to merge it before the #855?

void-mAlex: > Why do we need to merge it before the #855?

the idea is to not have a mega or that does everything



> @void-mAlex sorry I think I missed your point. This PR focuses solely on syntax change, without any functional changes.
> 
> Once we land this, we we would simply need to rebase #855, do all the necessary functional tweaks without noise of syntax conversion and land it with sole focus on v5 support via using e-engines-router-service

my concern is that a decent amount of time will be wasted debugging tests again, which is fine, it's not too daunting, but tried leaving a breadcrumb to one conflict I know we'll have

SergeAstapov: > tried leaving a breadcrumb to one conflict I know we'll have

noted! will be glad to help

PR #869, opened 2024-03-25
Add more LTS version to CI matrix

SergeAstapov:
another piece of overdue maintenance

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,9 @@ jobs:
           - ember-lts-3.24
           - ember-lts-3.28
           - ember-lts-4.4
+          - ember-lts-4.8
+          - ember-lts-4.12
+          # - ember-lts-5.4
           # - ember-release
           # - ember-beta
           # - ember-canary
--- a/packages/ember-engines/config/ember-try.js
+++ b/packages/ember-engines/config/ember-try.js
@@ -46,6 +46,17 @@ module.exports = async function () {
         }
       },
       {
+        name: "ember-lts-4.12",
+        npm: {
+          devDependencies: {
+            "ember-source": "~4.8.0",
+            "ember-cli": "~3.28.0",
+            "ember-cli-app-version": "^5.0.0",
+            "@ember/legacy-built-in-components": "~0.5.0",
+          }
+        }
+      },
+      {
         name: "ember-lts-5.4",
         npm: {
           devDependencies: {


packages/ember-engines/config/ember-try.js
```diff
@@ -45,6 +45,17 @@ module.exports = async function () {
...
         }
       },
+      {
+        name: "ember-lts-4.12",
+        npm: {
+          devDependencies: {
+            "ember-source": "~4.8.0",
+            "ember-cli": "~3.28.0",
+            "ember-cli-app-version": "^5.0.0",
+            "@ember/legacy-built-in-components": "~0.4.0",
```
void-mAlex, line 55: starting from this version we really should use latest legacy built in components 0.5.x which has a peer on this ember source

.github/workflows/ci.yml
```diff
@@ -54,6 +54,9 @@ jobs:
           - ember-lts-3.24
           - ember-lts-3.28
           - ember-lts-4.4
+          - ember-lts-4.8
```
void-mAlex, line 57: not sure what the value of keeping 4.4 around is

SergeAstapov: my thought process is that it should be each LTS from oldest supported

packages/ember-engines/config/ember-try.js
SergeAstapov: good point! Thanks!

.github/workflows/ci.yml
void-mAlex: ci isn't that slow now and this only adds some extra work which is what I'm concerned about, but it's probably fine

villander: I prefer adding the latest embroider to it
```suggestion
          - ember-lts-4.12
          - ember-lts-5.4
          - ember-release
          - ember-beta
          - ember-canary
          - embroider-safe
          - embroider-optimized
```

villander (changes requested):
We need to make it more realistic with ember status quo

.github/workflows/ci.yml
SergeAstapov: @villander v5 and embroider are omit intentionally, until we get #855 over the finish line

villander: @SergeAstapov can we get the finish line before change the pipeline? I’m pretty sure it has more priority than internal things

void-mAlex: 855 adds the extra pipelines you're talking about @villander, this bit is fine as is

void-mAlex:
this pr is ready, we're just holding off till release plan is set up so we benefit from automation for it

.github/workflows/ci.yml
villander: @void-mAlex thanks for the clarifications

SergeAstapov:
as #872 landed, merging this to test out release-plan cc @void-mAlex

PR #870, opened 2024-03-25
Drop support for Ember.js < 3.28

SergeAstapov:
As we need to have a new major anyways, it's good time to chip away real old version of Ember.js which could make support for green CI harder.

Versions 3.25-3.28 came out with many niceties.
At the same time, users who still can't bump to v4 for whatever reason can use 3.28 easily.


villander (approved):
Fair enough ✅

PR #872, opened 2024-03-25
Setup release-plan

SergeAstapov:
created using https://github.com/embroider-build/create-release-plan-setup

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "release-plan": "^0.8.0"
+    "release-plan": "^0.9.0"
   },
   "engines": {
     "node": "16.* || 18.* || >= 20"
--- a/package.json
+++ b/package.json
@@ -25,10 +25,9 @@
   "prettier": {
     "singleQuote": true
   },
+  "dependencies": {},
   "devDependencies": {
-    "@release-it-plugins/lerna-changelog": "^5.0.0",
-    "@release-it-plugins/workspaces": "^3.2.0",
-    "release-it": "^15.5.0"
+    "release-plan": "^0.8.0"
   },
   "engines": {
     "node": "16.* || 18.* || >= 20"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -25,14 +25,20 @@
   "prettier": {
     "singleQuote": true
   },
+  "dependencies": {},
   "devDependencies": {
     "@release-it-plugins/lerna-changelog": "^5.0.0",
     "@release-it-plugins/workspaces": "^3.2.0",
-    "release-it": "^15.5.0"
+    "release-it": "^15.5.0",
```
void-mAlex, line 32: do we still require release it?

package.json
```diff
@@ -25,14 +25,20 @@
...
+    "release-plan": "^0.8.0"
   },
   "engines": {
     "node": "16.* || 18.* || >= 20"
   },
+  "volta": {
+    "node": "16.20.2",
+    "yarn": "1.22.17"
```
void-mAlex, line 40: this seems wrong since we switched to pnpm

package.json
SergeAstapov: nope, thanks for noticing. removed!

package.json
SergeAstapov: indeed! removed

villander (changes requested):
I don't think we should switch our release workflow. it has been working well and I don't see any advantage using the release plan over what we have in place

SergeAstapov:
@villander `release-plan` does bit more automation compared to `release-it`, e.g. release preview PR so worth the shot (if we can make it work reliably easily)

void-mAlex:
> I don't think we should switch our release workflow. it has been working well and I don't see any advantage in using the release plan over what we have in place @SergeAstapov @void-mAlex

the issue with the current release flow is that it's not possible, due to mainly no npm access to the package

the change to release-plan ensures that npm access can be given only as a GH secret and access managed on GH with correct perissions to merge, it also ensures the builds are correct and reproducible with no human error 'possible' in the process as the only thing it involves are correct pr titles and labels
at the same time we get more automation and control over the release and if there is something else we need release plan to do it's easier to upstream as long as we have the use case.

for those reasons but mainly the ability to centrally release are why we need this before we merge the rest

villander:
thanks for the context @void-mAlex that makes sense now. We can move forward @SergeAstapov given the release has been a pain point to us requiring Chris to publish npm package every single time

villander (approved):

SergeAstapov:
Thank you @void-mAlex @mansona !

Issue #875, opened 2024-04-11
Memory leak in tests when using setupEngine helper


brettburley:
When using the newer `setupEngine` helper in a QUnit test, it appears that the engine instance and its container can be leaked after the test. This happens, for example, when using the ember-data addon which registers a test waiter, which is never cleaned up because `destroy` is never invoked on the engine.

When using the `setupEngine` helper, two contexts are created for most tests--one by `ember-test-helpers` for the dummy application used in tests, and then a second one that is created by the `setupEngine` call for the engine: https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/addon-test-support/index.js#L17.

Here's an example heapdump for a test module with 6 tests, where many instances of the ember-data store and respective containers are retained after the test runs:
<img width="1917" alt="Screenshot 2024-04-11 at 10 57 36 AM" src="https://github.com/ember-engines/ember-engines/assets/8064362/9502e321-ec34-4f08-a400-b16b358b2e42">

I am able to resolve the leak by registering the engine to be destroyed with the parent context used in the test by adding:
```
hooks.beforeEach(function() {
  associateDestroyableChild(this, this.engine);
});
```

This ensures that the engine's destroy hook is invoked when ember-test-helpers destroys the context and mirrors how the original `owner` is registered: https://github.com/emberjs/ember-test-helpers/blob/master/addon/addon-test-support/%40ember/test-helpers/setup-context.ts#L425. However, I'm not sure whether this registration should be part of the `setupEngine` method so that it's only used in tests, or whether there's some other relationship that should be defined as part of the engine instantiation in the core ember.js/ember-engines code.

SergeAstapov:
@brettburley Thank you for detailed report! It looks like to me `setupEngine` is the proper place for that line of code as in tests we create engine instance on [this line](https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/addon-test-support/index.js#L61) and then it's our responsibility to do the clean up.

doing `associateDestroyableChild(this, this.engine);` after we created `engineInstance` seems correct fix.

SergeAstapov:
closed by #876

PR #876, opened 2024-04-11
Ensure that the engine is destroyed in tests when the test context is cleaned up.

brettburley:
If the engine is not registered to be destroyed, instances of the engine can be leaked as reported in https://github.com/ember-engines/ember-engines/issues/875.


packages/ember-engines/tests/integration/setup-engine-test.js
```diff
@@ -16,6 +17,14 @@ module('Integration | setupEngine', function () {
       let route = this.engine.lookup('route:post.likes');
       assert.ok(route);
     });
+
+    test('it is destroyed with the test context', async function (assert) {
```
SergeAstapov, line 21: nice!

SergeAstapov (approved):

SergeAstapov:
@brettburley thank you!

PR #883, opened 2024-04-13
Drop support for Node.js v16

SergeAstapov:
this should be no possible after #882


SergeAstapov:
actually root cause is around tests setup and use of `ember-cli-addon-tests` for build tests

PR #885, opened 2024-04-13
run `npx ember-cli-update --to=5.7.0` to align with blueprint

SergeAstapov:


.github/workflows/ci.yml
```diff
@@ -59,17 +74,20 @@ jobs:
           # - ember-release
           # - ember-beta
           # - ember-canary
+          # - embroider-safe
```
SergeAstapov, line 77: we'll get to enabling these separately

PR #887, opened 2024-04-14
Implement {{lazy-mount}} component

SergeAstapov:
Extracted from https://github.com/buschtoens/ember-lazy-mount and slightly modernized.

Credit to @buschtoens for great work.

### DIFF
--- a/packages/ember-engines/addon/components/lazy-mount.hbs
+++ b/packages/ember-engines/addon/components/lazy-mount.hbs
@@ -1,0 +1,7 @@
+{{#if this.isLoading}}
+  {{yield (hash isLoading=true error=null)}}
+{{else if this.error}}
+  {{yield (hash isLoading=false error=this.error)}}
+{{else}}
+  {{mount this.loadedName model=@model}}
+{{/if}}
--- a/packages/ember-engines/addon/components/lazy-mount.js
+++ b/packages/ember-engines/addon/components/lazy-mount.js
@@ -1,0 +1,253 @@
+// eslint-disable-next-line ember/no-classic-components
+import Component from '@ember/component';
+import { tracked } from '@glimmer/tracking';
+import { assert } from '@ember/debug';
+import { inject as service } from '@ember/service';
+import { isDestroyed, isDestroying } from '@ember/destroyable';
+import {
+  dependencySatisfies,
+  importSync,
+  macroCondition,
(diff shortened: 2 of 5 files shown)


SergeAstapov:
IP because needs tests.

gossi:
I have a bit more modern code lying around from my migration. I will share it next week.

SergeAstapov:
discussed this yesterday with @void-mAlex and this is not future proof and won't work under Embroider.
This solution would work only under classic Ember CLI build.

We should think about what this would look like under Embroider so we have path forward.

villander:
@SergeAstapov @void-mAlex can we have a meeting to discuss it further? I'd love to release it given it's part of the original RFC and I can help making it work with Embroider + adding tests

gossi:
I am sending my changes in #892 with the idea of them being merged into this PR and then this will become one PR.

I made the discovery of what API design this could land on (component vs. helper), but I don't have an opinion and as I saw @villander [comment](https://github.com/ember-engines/ember-engines/pull/887#issuecomment-2083827273), I don't even have the reference to the RFC. Please make a decision, take the parts that are relevant and feel free to close my PR afterwards 👍

void-mAlex:
brought this up during the embroider team meeting and will attempt to summarize the discussion

there currently isn't a clear way on how to get this to work with optimized emberoider scenarios
this sort of feature would fall out naturally out of the v2 app format, but the pre-reqs for that are still being worked on
the recommendation would be to get involved with [v2 app rfc](https://github.com/emberjs/rfcs/pull/977) 

as an alternative if services isolation is not an issue, the lazy loading aspect can be achieved in the form of converting the engine to a strict template v2 addon format and use existing lazy loading component techniques to achieve part of the benefits from this component/helper

that's it for the summary, left a comment on the linked pr as well regarding dynamic imports

if people want to invest time into something that is going to work only on classic builds or some non-optimised embroider scenarios (little benefit to be had from those) they are free to pursue this feature

please note this work would need to land after ember source v5 compatibility lands together with the full ember try scenarios for it as well as embroider and would require all of those to pass in order to ensure users that are on those versions/setups would continue to work with the new release

villander:
@SergeAstapov would you mind to rebase it and see how it's fit into the embroider world? As @void-mAlex raised, it can be a block, so I'm more than happy to see what come next and make it compatible to move forward.

Issue #674, opened 2019-08-02
Instance Initializers not run before Integration Tests


amk221:
In an Ember Application, its Instance Initializers are run before each Component Integration Test.

In an Ember Engine, its Instance Initializers are not run before each Component Integration Test.

rwjblue:
Will be addressed by https://github.com/ember-engines/ember-engines/pull/653

amk221:
Ah, amazing, thanks

PR #888, opened 2024-04-14
[DEPRECATION] Deprecate transition methods of controller and route per RFC #674

SergeAstapov:
Deprecate transition methods to match https://github.com/emberjs/rfcs/pull/674 before we remove them in #855

this will be last thing to release in v0.10.0 before we go v1.0.0

TODO:
- [x] to add test cases similar to https://github.com/ember-engines/ember-engines/pull/764/files#diff-7cd3f1c5d330213360758d46650d5c63ca40826002621a02c7f1195a6321459fR119-R151
- [x] open PR to ember-engines.com with deprecation

### DIFF
--- a/packages/ember-engines/addon/-private/controller-ext.js
+++ b/packages/ember-engines/addon/-private/controller-ext.js
@@ -1,11 +1,14 @@
 import Controller from '@ember/controller';
 import { getOwner } from '@ember/application';
+import { deprecateTransitionMethods } from './deprecate-transition-methods';
 
 Controller.reopen({
   /*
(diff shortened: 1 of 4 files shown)


void-mAlex (commented):
assuming that is the correct place to get deprecate function from, which imo is a slightly weird place for it to come from,
this is LGTM

void-mAlex (approved):

SergeAstapov:
@void-mAlex @villander deprecation docs part was been completed https://ember-engines.com/docs/testing-integration

villander (approved):
Looks great!

PR #892, opened 2024-04-30
Earlier `<LazyMount>` migrations

gossi:
Heyo 👋 

This was my original `<LazyMount>` migration, to support embroider and classic builds. I modernized the component with a glimmer component (such no positional params) but named blocks and updated the loading of the engine with something that works under embroider as well as requirejs land.

I think, the loading part is the core of this PR.

While doing so and looking at the "modernizations" to `<LazyMount>` that sorta diverge from `{{mount}}` - I think _it might_ be better to revert this `<LazyMount>` component back into a `{{lazy-mount}}` helper.

Either way should be fine, but keeping the helper as an ember component is probably the worst decision to be made here.

### DIFF
--- a/packages/ember-engines/addon/services/engine-loader.js
+++ b/packages/ember-engines/addon/services/engine-loader.js
@@ -1,20 +1,15 @@
 import Service from '@ember/service';
-import require from 'require';
 import {
   dependencySatisfies,
   importSync,
   macroCondition,
 } from '@embroider/macros';
 
-let getOwner;
-
-if (macroCondition(dependencySatisfies('ember-source', '>=4.10'))) {
-  const { getOwner: _getOwner } = importSync('@ember/owner');
-  getOwner = _getOwner;
-} else {
-  const { getOwner: _getOwner } = importSync('@ember/application');
-  getOwner = _getOwner;
-}
+const { getOwner } = (
+  macroCondition(dependencySatisfies('ember-source', '>=4.10'))
+    ? importSync('@ember/owner')
+    : importSync('@ember/application')
+)
 
 export default class EngineLoader extends Service {
   /**
@@ -38,11 +33,46 @@ export default class EngineLoader extends Service {
    *
    * @param {String} name
    */
-  register(name) {
+  async register(name) {
     if (this.isLoaded(name)) return;
 
-    const owner = getOwner(this);
-    owner.register(`engine:${name}`, require(`${name}/engine`).default);
+    let module, engine;
+
+    // now here it gets a little tricky. The dynamic `import()` is ambivalent.
+    //
+    // 1) When webpack is used for the build system of the host app, it gets
+    //    rewritten to `__webpack_require__()`
+    //
+    // 2) In the classic build... well, I don't know?
+    //
+    // On the other hand is the `AssetLoaderService` from `ember-asset-loader`
+    // who can lazy load bundles, but that would end up in `requirejs` and
+    // `__webpack_require__` is not in sync with `requirejs`.
+    //
+    // So the ambivalent `import()` will under a classic build find the asset in
+    // the `requirejs` bundles but under a webpack build will throw an error
+    // (`Cannot find module \`${name}/engine\``).
+    //
+    // So here is the solution:
+    // 1. Try to use `import()` because it should be the go to API to use for
+    //    dynamically importing modules
+    // 2. When that fails, try to access `requirejs` directly.
+    //
+    try {
+      module = await import(`${name}/engine`);
+
+      engine = module.default;
+    } catch {
+      if (!engine && requirejs !== undefined) {
+        module = requirejs(`${name}/engine`);
+
+        engine = module.default;
+      }
+    }
+
+    if (engine) {
+      getOwner(this).register(`engine:${name}`, engine);
+    }
   }
 
   /**
--- a/packages/ember-engines/addon/components/lazy-mount.hbs
+++ b/packages/ember-engines/addon/components/lazy-mount.hbs
@@ -1,7 +1,9 @@
+{{this.load @name}}
+
 {{#if this.isLoading}}
-  {{yield (hash isLoading=true error=null)}}
+  {{yield to="loading"}}
 {{else if this.error}}
-  {{yield (hash isLoading=false error=this.error)}}
-{{else}}
+  {{yield this.error to="error"}}
+{{else if this.loadedName}}
   {{mount this.loadedName model=@model}}
 {{/if}}
--- a/packages/ember-engines/addon/components/lazy-mount.js
+++ b/packages/ember-engines/addon/components/lazy-mount.js
@@ -1,5 +1,5 @@
-// eslint-disable-next-line ember/no-classic-components
-import Component from '@ember/component';
+import Component from '@glimmer/component';
+import { next } from '@ember/runloop';
 import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
 import { inject as service } from '@ember/service';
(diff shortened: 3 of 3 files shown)


packages/ember-engines/addon/services/engine-loader.js
```diff
@@ -38,11 +33,46 @@ export default class EngineLoader extends Service {
...
+    // the `requirejs` bundles but under a webpack build will throw an error
+    // (`Cannot find module \`${name}/engine\``).
+    //
+    // So here is the solution:
+    // 1. Try to use `import()` because it should be the go to API to use for
+    //    dynamically importing modules
+    // 2. When that fails, try to access `requirejs` directly.
+    //
+    try {
+      module = await import(`${name}/engine`);
```
void-mAlex, line 62: this sort of import is 'illegal' under embroider as there is no way to know what it could refer to

villander:
@gossi could you please add tests to it including embroider scenario?

packages/ember-engines/addon/services/engine-loader.js
SergeAstapov: to me, this is kinda impossible to do reliably both for classic Ember CLI build *and* Enbroider without knowing how Embroider would handle lazy-loaded non-routable engine.

on the other hand, we could have "classic" only implementation like we have for lazy-loaded routable engine being provided by the `ember-engines` package itself (which would not "just work" under Embroider).
Then, in Embroider we would have `[compat-adapter](https://github.com/embroider-build/embroider/tree/main/packages/compat/src/compat-adapters)` that override that behavior, once Embroider can handle lazy-loaded non-routable engine.

As I understand after discussion with @void-mAlex, second part of the above currently impossible because Embroider not able to handle lazy-loaded non-routable engine.

And here we get to the gray area, because there is no clear understanding how lazy-loaded non-routable engine should be handled by Embroider, except understanding that Engines in future should be v2 apps per https://github.com/emberjs/rfcs/pull/977 (which is early draft as of writing this).
In that future, v2 apps could be built, deployed and connected together more freely as each app could be an engine or a host app but that's kinda more like an implementation detail.

Unfortunately, this all of the above does not give any answers, only asks more questions. But idea is to outline an understanding of current state of things.

And knowing above, I think it's still valuable to ship this feature knowing it's limited to classic Ember CLI build system because at this point in time we can't do other way.

In parallel, I don't mind to help push https://github.com/emberjs/rfcs/pull/977 if I could as it would benefit both Ember Engines project and wider ecosystem (and I assume would fit Polar goals too).

void-mAlex:
> @gossi could you please add tests to it including embroider scenario?

embroider scenario will be added as part of the ember v5 work
there are no currently viable solutions to make this work under embroider so the tests which are highly desirable would have to be skipped for those ember try scenarios

gossi:
The idea of this PR was to push my earlier investigations, as such the PR is not targeted at the `main` branch, but the `lazy-mount` branch.

That is, take from my PR what you need: Merge it into or copy the relevant lines and close it afterwards.

I expect final implementation and testing to happen at #887

What I could offer though: I do have the component and the service available as (g)ts 😬

packages/ember-engines/addon/services/engine-loader.js
gossi: I dunno if that was illegal, but nevertheless aware this is "grey area" here. However, as long as it allows to have engines being loaded under embroider this is a win (and then the crime doesn't really matter that much).

I also expect, that once v2 app RFC is more clear, that engines will be rewritten, ie. drop broccoli for unplugin etc. which then would also solve the crime that was lasting long enough to enable people to update to embroider.

So, as long as the consumer API is "stable", the implementation behind the curtains can change and update along the versions.

gossi: The situation to avoid is to wait for app v2 RFC to be finished and _then_ make engines embroider compatible... that'll take too long.

void-mAlex:
@gossi curious which ember source you're on and which embroider flags you have one + embroider version
the code on this branch is likely incidentally working because of requirejs being available under certain versions/configurations of ember-source/embroider
even with gts it will only work by requirejs existing and having the correct entries

the import statement does not work now and will never work in the future as written no matter the build tool used
```import(`${dynamicSegmentFirst}/something`)``` is an error no matter where you try to use it.
only something like ```import(`known/static/path/first/${dynamicSegment}/something/something`)``` could technically work

**Comment by gossi** [comment]
Ah, I did use the addon v2 blueprint and used the test-app within there (with `"ember-source": "~5.7.0"` as dep).

In `ember-cli-build.js` I was trying these both approaches:

```ts
  // const { maybeEmbroider } = require('@embroider/test-setup');
  // return maybeEmbroider(app);

  return app.toTree();
```

But I see, it only has `"@embroider/test-setup": "^3.0.1"` as dev-dependency, so the `maybeEmbroider()` afaik returns false then, so no embroider, but webpack.

packages/ember-engines/addon/services/engine-loader.js
void-mAlex: > I also expect, that once v2 app RFC is more clear, that engines will be rewritten, ie. drop broccoli for unplugin etc. which then would also solve the crime that was lasting long enough to enable people to update to embroider.

people can upgrade to embroider today with no issues, so I'd like to clear up that as it feels like that was a misunderstanding somewhere. I'm on latest stable embroider with all static flags on and using lazy routable engines.

this feature would not work though in any scenario where embroider would bring you benefits though and would actually block people from adopting if they tried to use it.

> The situation to avoid is to wait for app v2 RFC to be finished and _then_ make engines embroider compatible... that'll take too long.

v2 app rfc needs to progress further than it is now and then there will be no more need for engines package at all, no need for an unplugin or anything of the sorts

the stable api that is needed for engines is how it deals with service isolation which is its own rfc that blocks the v2 app one and how it integrates with routing with is partially blocking the v2 app rfc

villander: > The situation to avoid is to wait for app v2 RFC to be finished and then make engines embroider compatible... that'll take too long.

I couldn't agree more here @void-mAlex @SergeAstapov the [Ember Engines RFC build in 2014](https://rfcs.emberjs.com/id/0010-engines/) will complete 10 years and never were finished. I think is a tragedy to wait for something we're not really sure. At this point we need to complete the original RFC and wait for the new Engines that will come up with this new RFC. I love the way @void-mAlex is planning the new engines and I'll support it, but this time we need to be faithful to what we have today and do our best to make the current engines compatible with Embroider and finish the original RFC.

Please let's avoid overengineering here and make it simple to pay the whole tech debt that made the whole community frustrated and struggling with engines 🙏  


Additional Links:

https://discuss.emberjs.com/t/engines-1-0-roadmap/14914

gossi: I agree. This ain't be affecting public APIs (which is lazy mount) but is replaceable under the hood as embroider progresses and can be exchanged behind the curtains at a later point.

The more urgent question is will lazy mount be a helper or component, as this _is_ part of the public API. Once this is settled, I think all the bits and pieces are lying around. Copy from my PR what you need, close it and deliver the finishing blow over at #887 👍

void-mAlex: the only way the public api doesn't change is if it's the one used by the v2 app rfc (tbd at the point in time of this comment)
with that said, one thing to consider is for it to actually be a modifier as you'd technically need access to a dom node to render the engine in (which you would get with a component as well but not quite the same) between component and helper I'd got with component as we can do lazy loading today with that

I do want to stress that your public api will change if you want to support this feature under embroider/vite or whatever comes next

villander:
@gossi any update here?

gossi:
our solution was to turn (lazy)mount engines into (v2) addons and let ember's build system take control of chunking (and also such loading). I stopped this PR then.

Probably I vote for deprecating them. However for continuing support they need to be migrated. Feel free to take over my PR. Consider this a PoC and start over, contribute to that, etc.

villander:
> Probably I vote for deprecating them

Sorry, I think I didn't get your point. Why is it deprecated, if **lazy-mount** is a required feature for routeless engines to be lazy loading?

PR #902, opened 2025-03-11
update release-plan

mansona:


patches/release-plan@0.9.0.patch
```diff
@@ -1,30 +0,0 @@
...
-                     impact,
-                     oldVersion: entry.version,
-@@ -51,6 +51,16 @@ class Plan {
-         }
-         return solution;
-     }
-+    #configureImpact(pkgName, impact) {
-+        const packageJson = this.#pkgs.get(pkgName)?.pkg;
-+        if (packageJson && packageJson['release-plan']?.semverIncrementAs) {
-+            const semverOverrides = packageJson['release-plan'].semverIncrementAs;
```
SergeAstapov, line 21: @mansona could the new version handle pre-release versions? any setup we can/need to update?

void-mAlex: the patch got upstreamed and the config is the same

mansona: yes @void-mAlex is right, the fix has been upstreamed and you will see the effect once this is merged i.e. you will see the Plan release PR before anything will get released and it will say what version it's about to release 👍

.github/workflows/publish.yml
```diff
@@ -10,54 +9,35 @@ on:
...
-      # plus the fact that this action only runs on main will be enough of a guard
-      - id: check-release
-        run: if git diff --name-only HEAD HEAD~1 | grep -w -q ".release-plan.json"; then echo "command=release"; fi >> $GITHUB_OUTPUT
-
   publish:
     name: "NPM Publish"
     runs-on: ubuntu-latest
-    needs: check-plan
-    if: needs.check-plan.outputs.command == 'release'
-    environment: deploy
```
void-mAlex, line 41: believe you still need this to have access to the npm token

void-mAlex:
@SergeAstapov  I'm thinking we merge this once we get environment back in the ci file then see tests passing on #903

.github/workflows/publish.yml
mansona: ah yes I don't use environments so I wasn't aware 👍

void-mAlex (approved):

PR #904, opened 2025-03-11
fix action deprecation

mansona:
This PR extracts the one commit from #903 that actually fixes the problem. It looks like we're on a bit of a cliff with this project where we need old versions of things that just don't work when you upgrade them 🙈 probably something to discuss

the floating dependencies fail will be fixed by #902 👍


void-mAlex (approved):

void-mAlex:
the old things are 're-introduced' by an addon

SergeAstapov (approved):

PR #905, opened 2025-03-12
Move from using ember-cli-addon-tests to using fixturify-project for tests

mansona:
This should unblock #903 and is a better reflection of how people will be consuming the addon 👍 

Edit: oh and this almost halves the CI time for the node tests 🎉


SergeAstapov:
Thank you @mansona! this could be a great example to learn how to use `fixturify-project`

PR #907, opened 2025-04-13
Setup vite app tests

void-mAlex:


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 20
```
mansona, line 26: Can we keep this as 18 for the next 8 days 😉

packages/ember-engines/app/engine.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ember-engines/engine';
```
mansona, line 1: do we need to do an app export here? I assume we can rely on addon exports properly if it's intended to be imported explicitly?

packages/ember-engines/package.json
```diff
@@ -125,8 +125,7 @@
     "lodash": "^4.17.11"
   },
   "peerDependencies": {
-    "ember-engines-router-service": ">=0.5.0",
-    "ember-source": "^3.28 || >= 4.0.0"
```
mansona, line 129: complete removal of ember-source as a peer? seems... scary 😂 maybe just expand the range?

packages/vite-app/vite.config.mjs
```diff
@@ -0,0 +1,23 @@
...
+      babelHelpers: 'runtime',
+      extensions,
+    }),
+  ],
+  // build: {
+  //   target: "ES2022"
+  // },
+  optimizeDeps: {
+    esbuildOptions:{
+      target: 'ES2022'
```
mansona, line 20: we talked about adding this upstream btw, to support the exact use case that you're targeting 👍

patches/ember-lifeline@7.0.0.patch
```diff
@@ -0,0 +1,19 @@
...
+-    "@ember/test-helpers": ">= 1.0.0"
+-  },
+-  "peerDependenciesMeta": {
+-    "@ember/test-helpers": {
+-      "optional": true
+-    }
+-  },
+   "resolutions": {
+     "favicons": "5.3.0"
+   },
```
mansona, line 19: I assume this is a WIP since it's not active in the package.json? 🤔

patches/tracked-toolbox@2.0.0.patch
```diff
@@ -0,0 +1,19 @@
...
+-    "ember-source": "*"
+-  },
+-  "peerDependenciesMeta": {
+-    "ember-source": {
+-      "optional": true
+-    }
+-  },
+   "engines": {
+     "node": "14.* || 16.* || >= 18"
+   },
```
mansona, line 19: same question as above

mansona (commented):
this is looking real nice 😍

.github/workflows/ci.yml
void-mAlex: won't get merged in the next 8 days :)

